PR #10, opened 2024-04-04
Add simulate-sce module

jashapiro:
This draft PR starts to add a `simulate-sce` module. 

The R scripts are derived from https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/63f6d91464ef86bf15b8c2e9c7e56f048e4c1a9b/analyses/simulate-sce, with additional nextflow wrapping.

The main processes are pretty straightforward, though I used the `stageAs` directive a few times to simplify managing files that will have the same names for input and output.

I also implemented a few `Util` functions, namely to get lists of projects and project paths from an S3 release bucket. 

Note that the way this is implemented is a divergence from my initial idea: I now think we will likely want to have some standard input channels to subworkflows, which may be just a channel of `[project_id, project_path]` as I have done here. 

While I initially thought I would have the whole project directory as inputs to processes themselves, I quickly realized that this resulted in lots of extra staged files, and I wanted to be able to work in parallel by sample, so I also implemented some channel logic to create sets of input files for each sample in a project. I expect that some of this will be generally useful as well, so I expect to move some part or parts of it to `Utils`.

The idea would be to have custom functions that take a channel of Projects and emit channels with specific subsets of the files: most likely with only processed rds or hdf5 files. I think that would happen in a later PR, and maybe after we have a few modules implemented to see what patterns emerge. 

Right now this workflow works locally with a local R install, but the Docker image specified is not actually not correct, as it does not have `splatter` installed (and it has a lot of extras!). I plan to make a proper Docker image in openscpca-analysis, then we can use that here. But I thought that I would get this draft up for a look before I went _too_ far along.


main.nf
```diff
@@ -2,15 +2,34 @@
...
 if (param_error) {
   System.exit(1)
 }
 
 // **** Main workflow ****
 workflow {
   example()
+  // project channel of [project_name, project_path]
+  project_ch = Channel.fromPath(Utils.getProjectPaths(release_dir))
+    .map{[it.name, it]}
```
allyhawkins, line 32: Is the project name going to be SCPCPX by using `it.name`?

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,103 @@
...
+process permute_metadata {
+  container 'ghcr.io/alexslemonade/scpca-tools:v0.3.2'
+  tag "$project_id"
+  publishDir "${params.sim_pubdir}/${project_id}", mode: 'copy'
+  input:
+    tuple val(project_id),
+          path(metadata_file, stageAs: 'input/*')
+  output:
+    tuple val(project_id), path("${metadata_file.fileName.name}")
+  script:
```
allyhawkins, line 12:
```suggestion
    tuple val(project_id), path("${permuted_metadata}")
  script:
  permuted_metadata = "${metadata_file.fileName.name}"
```
I feel like this is worth defining?

modules/simulate-sce/resources/usr/bin/permute-metadata.R
```diff
@@ -0,0 +1,134 @@
...
+    c("-f", "--metadata_file"),
+    type = "character",
+    default = NULL,
+    help = "Path to the sample metadata file to shuffle."
+  ),
+  make_option(
+    c("-o", "--output_file"),
+    type = "character",
+    default = NULL,
+    help = "The output file path."
```
allyhawkins, line 20: The output here is permuted metadata correct? I think that's worth noting in the help message and maybe in the argument name.

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,103 @@
...
+    metadata_ch = project_ch.map{[it[0], it[1] / 'single_cell_metadata.tsv']}
+    permuted_metadata_ch = permute_metadata(metadata_ch)
+
+    // get bulk files for each project, if present
+    bulk_ch = project_ch.map{[it[0], it[1] / 'bulk_quant.tsv', it[1] / 'bulk_metadata.tsv']}
+      .filter{it[1].exists()}
+    permute_bulk(bulk_ch)
+
+    // list rds files for each project and sample
+    sample_ch = project_ch.map{[it[0], it[1].listFiles().findAll{it.isDirectory()}]}
```
allyhawkins, line 93: This gives you a channel with [project id, sample directories] correct?

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,103 @@
...
+      .transpose() // transpose to get a channel of project ids and sample directories
+      // get rds file list for each sample
+      .map{[it[0], it[1].name, it[1].listFiles().findAll{it.name.endsWith(".rds")}]}
```
allyhawkins, line 96: And then this is giving you [project id, sample id, list of rds files for that sample]? I think I would put the structure in the comments since that can be helpful.

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,103 @@
...
+    """
+}
+
+process permute_bulk{
+  container 'ghcr.io/alexslemonade/scpca-tools:v0.3.2'
+  tag "$project_id"
+  publishDir "${params.sim_pubdir}/${project_id}", mode: 'copy'
+  input:
+    tuple val(project_id),
+          path(bulk_quant, stageAs: 'input/*'),
```
allyhawkins, line 61: Do you need the `stageAs` here if you only have one file that's being input? I think it's more helpful for when you have multiple files that will have the same filename, but there should only be one bulk file per project.

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,103 @@
+params.sim_pubdir = 's3://openscpca-temp-simdata/test'
+
+process permute_metadata {
+  container 'ghcr.io/alexslemonade/scpca-tools:v0.3.2'
+  tag "$project_id"
+  publishDir "${params.sim_pubdir}/${project_id}", mode: 'copy'
+  input:
+    tuple val(project_id),
+          path(metadata_file, stageAs: 'input/*')
```
allyhawkins, line 9: I think same thing as my comment on bulk that you don't really need this since there will only be one file per project? I thought you only had issues with conflicting file names if you have two files being read in the same process that were conflicting.

allyhawkins (commented):
I just had a few small questions, but this approach seems good to me.

main.nf
jashapiro: Yes, because the project directory is the `SCPCPXXX`

modules/simulate-sce/main.nf
jashapiro: I don't think it is a real issue, but I wanted to not have the output in a subdirectory, and it will have the same name as the input file.

modules/simulate-sce/main.nf
jashapiro: yeah, probably not a bad idea

modules/simulate-sce/main.nf
jashapiro: Yes, though I would style it as `[project id, [sample directories]]` or  `[project id, [sample_dir1, sample_dir2, sample_dir2, ...]]`

modules/simulate-sce/main.nf
jashapiro: Yes, I will add that comment. Meant to do it before review, but then wanted to get it up and forgot.

jashapiro (commented):
Thanks for the comments. I will update some of my own comments and respond more soon!

modules/simulate-sce/main.nf
allyhawkins: I thought it would let you just overwrite the input and that we had scenarios in `scpca-nf` where we modified the input by overwriting it, but I can't seem to find those. Probably better to do it this way anyways and not overwrite.

jashapiro:
I have now run this and with the added bug fixes, everything worked! A few random failures, as would be expected, but every sample now has simulated data. Ready for another look @allyhawkins

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,107 @@
+params.sim_pubdir = 's3://openscpca-temp-simdata/test'
+params.simulate_sce_container = 'ccdl/openscpca-simulate-sce:latest'
```
allyhawkins, line 2: Is the plan to always define module specific parameters in `main.nf` for that module?

modules/simulate-sce/main.nf
```diff
@@ -0,0 +1,107 @@
...
+    permuted_file = metadata_file.fileName.name
+    """
+    permute-metadata.R \
+      --metadata_file ${metadata_file} \
+      --output_file ${permuted_file}
+    """
+  stub:
+    permuted_metadata = metadata_file.fileName.name
+    """
+    touch ${permuted_file}
```
allyhawkins, line 23:
```suggestion
    permuted_file = metadata_file.fileName.name
    """
    touch ${permuted_file}
```

modules/simulate-sce/readme.md
```diff
@@ -0,0 +1,7 @@
+# Workflow to simulate single cell data
+
+This workflow is designed to simulate single cell data, prinically using the [splatter](https://bioconductor.org/packages/release/bioc/html/splatter.html) package.
```
allyhawkins, line 3:
```suggestion
This workflow is designed to simulate single cell data, primarily using the [splatter](https://bioconductor.org/packages/release/bioc/html/splatter.html) package.
```
Is this the word you meant to use?

modules/simulate-sce/readme.md
```diff
@@ -0,0 +1,7 @@
...
+
+Scripts are derived from the the `simulate-sce` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
+
+Permalink to the version used: https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/63f6d91464ef86bf15b8c2e9c7e56f048e4c1a9b/analyses/simulate-sce
```
allyhawkins, line 7: Maybe update this to include the latest changes you added in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/365

allyhawkins (approved):
LGTM, I just had a few minor comments but nothing I need to see again. 
I did not review the scripts line by line since I assumed you copied them over from the analysis directory, but let me know if you would like me to do that.

modules/simulate-sce/main.nf
jashapiro: I'm not sure, but it seemed a reasonable place to start.

PR #13, opened 2024-04-18
Require http tokens for metadata (IMDSv2)

jaclyn-taroni:
Right now, EC2 instances have been created with IMSDv2 as optional. We want to enforce that. Because we're using an Amazon Linux 2023 image, I would have expected us to get that without any intervention, but that is not what the console reports. Here's a link to the documentation I referenced: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#metadata-options


jashapiro (approved):
LGTM

PR #14, opened 2024-04-19
Use nextflow-specific key

jashapiro:
In continuing quest to get Batch running within the new infrastructure, I am here switching to using the queues we set up within the OpenScPCA account structure.

I also updated to use a KMS key for the batch instances that is managed in this repository, and moved the account number out to a variable to make changing it easier.

Unfortunately, we still seem to be missing something to allow the batch jobs to actually run: right now the instance launches but no jobs start from the queue

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,0 +1,2 @@
+[default.extend-words]
+kms = "kms"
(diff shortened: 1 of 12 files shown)


jaclyn-taroni (approved):
👍🏻

PR #17, opened 2024-04-25
Use h5ad extension for simulated data

### LINKED ISSUE #16: Use h5ad file extension
Right now we only have one module that does anything with AnnData files: the simulation script. This needs to be updated to output files with an`.h5ad` extension.

jashapiro:
Closes #16

Pretty small change, but needed before rerunning this!


sjspielman (approved):
I ran `grep -r "hdf5" .` in `modules/simulate-sce`, and nothing more came up 👍

PR #18, opened 2024-04-30
Updates for running in openscpca accounts

jashapiro:
This PR implements the changes to accounts that were required to get nextflow/batch up and running in the VPC account. 

These include:
- switching to the vpc and subnets defined in the workloads account
- add `ec2.amazonaws.com` to `openscpca-nf-ecs-instance-role`
- add the AmazonSSMManagedInstanceCore policy to `openscpca-nf-ecs-instance-role`

I also changed the output for the simulation module to be the final bucket, with the thought that I can just clone over to the temporary location later.

Finally (though this was done first), I added a profile that allows us to use the old batch queue. This is meant to be temporary, until we are confident that everything is working as expected!


jaclyn-taroni (approved):
This looks fine, but I'm not convinced the VPC that's not being used should remain for reference.

jashapiro:
> This looks fine, but I'm not convinced the VPC that's not being used should remain for reference.

That seems fair. I will remove it. We have an equivalent elsewhere if we need it.

PR #20, opened 2024-05-01
Set batch limit to quota

jashapiro:
Here I am updating the terraform to match the quota we were given by AWS for the workloads account.


jaclyn-taroni (approved):
👍🏻

PR #24, opened 2024-05-08
Add merge workflow

### LINKED ISSUE #6: Port merge workflow to OpenScPCA-nf
The project data release bucket will not include merged data, so we will need to port the `merge.nf` workflow from https://github.com/AlexsLemonade/scpca-nf to a module in this workflow. 

This will be a good test of the overall infrastructure we set up for this environment, as this workflow will require large amounts of disk space and high-memory instances.

jashapiro:
closes #6 

Here I am adding a module with the `merge_sce` workflow, ported from https://github.com/AlexsLemonade/scpca-nf.

As noted in the readme, I made a few changes along the way.
Most of these were simplify some assumptions and make the workflow work with the input of a list of project folders rather than from a sample sheet. 

One part of this meant that I had to figure out whether the adt was present for each sample, as we use that to determine a few things... I ended up doing this by looking for `_adt.h5ad` files, which seemed the simplest way to make sure we had that information, but I really only use this variable in the AnnData conversion. I probably could have incorporated something to make it work by further modifying the export script, but I did not do that as yet. (I did modify the `merge_sce.R` script to look for ADTs directly, mostly because the script was already doing that in another context.) 

Other than that, things do seem to run as expected. I tested it with `SCPCP000012`, and I am currently testing with `SCPCP000007` to make sure the ADT component works as expected. 

I am filing this as a draft partly because I want to make sure that the changes in progress to enable this workflow more automatically are not interrupted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Further instructions for this will be added in the future, and we expect this to
 The following base command will run the workflow, assuming all AWS permissions are set up correctly:
 
 ```bash
-nextflow run alexslemonade/openscpca-nf -profile batch
+nextflow run AlexsLemonade/OpenScPCA-nf -profile batch
 ```
 
 The workflow also has a couple of entry points other than the main workflow, for testing and creating simulated data.
@@ -27,7 +27,7 @@ nextflow run AlexsLemonade/OpenScPCA-nf -profile batch -entry test
 To run the workflow that creates simulated SCE objects for the OpenScPCA project, you can use the following command:
 
 ```bash
-nextflow run alexslemonade/openscpca-nf -profile batch -entry simulate
+nextflow run AlexsLemonade/OpenScPCA-nf -profile batch -entry simulate
 ```
 
 ## Repository setup
--- a/README.md
+++ b/README.md
@@ -18,10 +18,10 @@ nextflow run alexslemonade/openscpca-nf -profile batch
 
 The workflow also has a couple of entry points other than the main workflow, for testing and creating simulated data.
 
-To run a test version the workflow to check permissions and infrastructure setup:
+To run a test version of the workflow to check permissions and infrastructure setup:
 
 ```bash
-nextflow run alexslemonade/openscpca-nf -profile batch -entry test
+nextflow run AlexsLemonade/OpenScPCA-nf -profile batch -entry test
 ```
 
 To run the workflow that creates simulated SCE objects for the OpenScPCA project, you can use the following command:
--- a/README.md
+++ b/README.md
@@ -4,6 +4,32 @@ A workflow for running analysis modules ported from the [OpenScPCA-analysis repo
 
 See https://github.com/AlexsLemonade/OpenScPCA-admin/blob/main/technical-docs/nextflow-workflow-specifications.md for initial implementation plans for this workflow.
 
+## Running the workflow
+
+The workflow is currently set up to run best via AWS batch, but some testing may work locally.
+You will need to have appropriate AWS credentials set up to run the workflow on AWS and access the data files.
+Further instructions for this will be added in the future, and we expect this to be run via a GitHub Action for most use cases.
+
+The following base command will run the workflow, assuming all AWS permissions are set up correctly:
+
+```bash
+nextflow run alexslemonade/openscpca-nf -profile batch
+```
+
+The workflow also has a couple of entry points other than the main workflow, for testing and creating simulated data.
+
+To run a test version the workflow to check permissions and infrastructure setup:
+
+```bash
+nextflow run alexslemonade/openscpca-nf -profile batch -entry test
+```
+
+To run the workflow that creates simulated SCE objects for the OpenScPCA project, you can use the following command:
+
+```bash
+nextflow run alexslemonade/openscpca-nf -profile batch -entry simulate
+```
+
 ## Repository setup
 
 This repository uses [`pre-commit`](https://pre-commit.com) to enforce code style and formatting.
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -36,4 +36,27 @@ class Utils {
     def projects = getProjects(release_path)
     return projects.collect{release_path / it}
   }
+
+  static def getProjectFiles(project_path, format = "sce", process_level = "processed"){
+    project_path = Nextflow.file(project_path, type: 'dir')
+    process_level = process_level.toLowerCase()
+    if (!(process_level in ["raw", "filtered", "processed"])){
+      throw new IllegalArgumentException("Unknown process_level '${process_level}'")
+      }
+    def files = []
+    switch (format.toLowerCase()){
+      case ["anndata", "h5ad"]:
+        files = Nextflow.files(project_path / "**_${process_level}_*.h5ad")
+        break
+      case ["sce", "rds"]:
+        def extension="rds"
+        files = Nextflow.files(project_path / "**_${process_level}.rds")
+        break
+      default:
+        throw new IllegalArgumentException("Unknown format '${format}'")
+        break
+    }
+    files = files.findAll{it.size() > 0}
+    return files
+ }
 }
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -51,6 +51,7 @@ class Utils {
     def files = []
     switch (format.toLowerCase()){
       case ["anndata", "h5ad"]:
+        // find all h5ad files in the project directory (** searches all subdirectories)
         files = Nextflow.files(project_dir / "**_${process_level}_*.h5ad")
         break
       case ["sce", "rds"]:
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -6,7 +6,8 @@ import nextflow.Nextflow
  */
 class Utils {
   static def getReleasePath(bucket, release = "current"){
-    def bucket_path = Nextflow.file(bucket)
+
+    def bucket_path = Nextflow.file(bucket, type: 'dir')
     if (!bucket_path.exists()) {
       throw new IllegalArgumentException("Bucket '${bucket}' does not exist")
     }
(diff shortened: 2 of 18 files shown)


modules/merge-sce/main.nf
```diff
@@ -0,0 +1,161 @@
...
+        // multiplexed libraries are subdirectories with more than one sample id
+        multiplexed: files(it[1] / "*", type: "dir").any{it.name =~ /SCPCS\d+_SCPCS\d+/}
+        single_sample: true
+      }
+
+    // get all SCE files by project
+    // this will be a channel of [project_id, [library_ids], [processed_sce_files], has_adt]
+    libraries_ch = project_branch.single_sample
+      .map{project_id, project_dir -> {
+        def processed_files = files(project_dir / "**_processed.rds")
```
allyhawkins, line 127:
```suggestion
        def processed_files = files(project_dir / "**_processed.rds").size > 0
```
We should probably check the file size here too. Although they shouldn't exist in the release bucket, we don't want to accidentally read one in.

modules/merge-sce/main.nf
```diff
@@ -0,0 +1,161 @@
...
+        def library_ids = processed_files.collect{it.name.replace('_processed.rds', '')}
+        def has_adt = files(project_dir / "**_processed_adt.h5ad").size > 0 // true if there are any adt files
```
allyhawkins, line 129: I think for checking within Nextflow this is a fine approach. Alternatively, you could just leave this to the script itself and not worry about it here. 
I'm trying to think of a scenario where we would have CITE, but this file would be missing. I think the only time is if a library is also multiplexed, but that's not a thing we have encountered or should worry about.

modules/merge-sce/main.nf
```diff
@@ -0,0 +1,161 @@
...
+def publish_merge_base = "${params.results_bucket}/${params.release_prefix}/merge_sce"
+def merge_report_template = "${projectDir}/modules/merge-sce/resources/merge-report.rmd"
+
+
+// merge individual SCE objects into one SCE object
+process merge_group {
+  container "ghcr.io/alexslemonade/scpcatools-slim:edge"
+  tag "${merge_group_id}"
+  label 'mem_max'
+  label 'long_running'
```
allyhawkins, line 20: Do we have this label defined in this workflow? Looking at `process_base.config` I don't see it?

modules/merge-sce/main.nf
```diff
@@ -0,0 +1,161 @@
...
+    input_library_ids = library_ids.join(',')
+    input_sces = processed_files.join(',')
+    merged_sce_file = "${merge_group_id}_merged.rds"
+    """
+    merge_sces.R \
+      --input_library_ids '${input_library_ids}' \
+      --input_sce_files '${input_sces}' \
+      --output_sce_file '${merged_sce_file}' \
+      --n_hvg ${params.num_hvg} \
+      --threads ${task.cpus}
```
allyhawkins, line 36: I think you're missing use of the `has_adt` value.

modules/merge-sce/main.nf
```diff
@@ -0,0 +1,161 @@
...
+    merge_report = "${merge_group_id}_merged-summary-report.html"
+    """
+    touch ${merge_report}
+    """
+}
+
+process export_anndata {
+    container "ghcr.io/alexslemonade/scpcatools-anndata:edge"
+    label 'mem_max'
+    label 'long_running'
```
allyhawkins, line 80: Just noting this label again.

modules/merge-sce/main.nf
allyhawkins: Oh I see now that you are only using it to for exporting AnnData and not here.

modules/merge-sce/resources/usr/bin/merge_sces.R
```diff
@@ -0,0 +1,350 @@
...
+  purrr::map(\(sce) names(colData(altExp(sce, "adt")))) |>
+  # use .init to handle an empty input; will return the .init value
+  # if input is empty, then there are no "adt" altExps anyways
+  purrr::reduce(union, .init = NULL)
+
+# ensure that there are indeed no "adt" altExps if adt_present_columns is empty
+adt_present <- sce_list |>
+  purrr::map_lgl(\(sce) "adt" %in% altExpNames(sce))
+
+if (is.null(adt_present_columns) && any(adt_present)) {
```
allyhawkins, line 246:
```suggestion
# determine if adt is present for any SCE objects
adt_present <- sce_list |>
  purrr::map_lgl(\(sce) "adt" %in% altExpNames(sce))

# ensure that there are indeed no "adt" altExps if adt_present_columns is empty
if (is.null(adt_present_columns) && any(adt_present)) {
```
I think this comment might have been out of place?

modules/merge-sce/resources/usr/bin/merge_sces.R
```diff
@@ -0,0 +1,350 @@
...
+    help = "number of high variance genes to use for dimension reduction;
+            the default is n_hvg = 2000"
+  ),
+  make_option(
+    opt_str = c("--multiplexed"),
+    action = "store_true",
+    default = FALSE,
+    help = "Indicates if the provided SCE's contain multiplexed data.
+      If so, the sample metadata will not be added to the colData."
+  ),
```
allyhawkins, line 37: Do we still need this here?

allyhawkins (commented):
This mostly looks good to me. I don't have a lot of comments and I didn't go line by line on the template or the anndata scripts. I assumed they were copied over exactly so if you need me to take a closer look at those let me know. 

I think the only real comment I have that needs to be addressed is the presence of the `long_running` label. I think that's just missing from our configs?

modules/merge-sce/resources/usr/bin/merge_sces.R
jashapiro: I am not sure, but I didn't want to remove it in case we want it later...

modules/merge-sce/main.nf
jashapiro: Doing this a slightly different way: what you have would check that the array is not empty, but it would not check the individual elements.

modules/merge-sce/main.nf
jashapiro: added it to the batch config

jashapiro:
Since you last looked @allyhawkins, I added a function to get the files for a given project in a similar way to the download script. I expect to use this in some future updates, in particular when we might link up the simulation module to run through the rest of the workflow. This should remove some boilerplate in later modules in any case.

I also removed the `has_adt` value completely. The only real problem with this is that the stub workflow for moving AnnData doesn't know whether it has to make an `_adt` file. Which really should not be a big problem.

I'm still testing a bit, but the big hurdle was overcome, which was learning how Groovy handles named arguments. I can not emphasize how much they do not work the way I expected them to work.

lib/Utils.groovy
```diff
@@ -35,4 +37,31 @@ class Utils {
...
+    project_dir = Nextflow.file(project_dir, type: 'dir')
+    println(project_dir)
+    process_level = process_level.toLowerCase()
+    if (!(process_level in ["raw", "filtered", "processed"])){
+      throw new IllegalArgumentException("Unknown process_level '${process_level}'")
+      }
+    def files = []
+    switch (format.toLowerCase()){
+      case ["anndata", "h5ad"]:
+        files = Nextflow.files(project_dir / "**_${process_level}_*.h5ad")
```
allyhawkins, line 54: Does this recursively search within nested directories in the `project_dir`, I assume yes?

modules/merge-sce/readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+The workflow and scripts are largely ported from the [`scpca-nf` workflow](https://github.com/AlexsLemonade/scpca-nf).
+The [a9dc826 commit](https://github.com/AlexsLemonade/scpca-nf/tree/a9dc826b8576c48ca38c6ca137f9eeced29c3acc) was used as the base for the port.
+
+There have been some small changes, in particular:
+
+- Containers have been updated to use a more recent version of `scpcaTools`, with underlying updates to Bioconductor and Python packages.
+- The workflow is run on a project level, and all SCE files within a project are merged.
+  - the assumption in this workflow is that all libraries within a project are to be merged, so there (currently) are no options to specify which specific samples to merge.
+- Presence of ADTs is determined within the workflow by the presence of `_adt.h5ad` files among the sample files.
```
allyhawkins, line 13: I don't think this is true anymore?

lib/Utils.groovy
allyhawkins: Also can you maybe add some comments in here?

allyhawkins (commented):
I think you just need to update some comments and documentation, but other than that looks good.

jashapiro:
Okay, I think this is ready to go in. The last thing I did was add some separate entrypoints to the main workflow, so we can run just the test or simulation portions separately if we want.

README.md
```diff
@@ -4,6 +4,32 @@ A workflow for running analysis modules ported from the [OpenScPCA-analysis repo
...
+
+The following base command will run the workflow, assuming all AWS permissions are set up correctly:
+
+```bash
+nextflow run alexslemonade/openscpca-nf -profile batch
+```
+
+The workflow also has a couple of entry points other than the main workflow, for testing and creating simulated data.
+
+To run a test version the workflow to check permissions and infrastructure setup:
```
allyhawkins, line 21:
```suggestion
To run a test version of the workflow to check permissions and infrastructure setup:
```

README.md
```diff
@@ -4,6 +4,32 @@ A workflow for running analysis modules ported from the [OpenScPCA-analysis repo
...
+
+```bash
+nextflow run alexslemonade/openscpca-nf -profile batch -entry test
```
allyhawkins, line 24:
```suggestion
nextflow run AlexsLemonade/OpenScPCA-nf -profile batch -entry test
```
At least on my end, I can't pull the workflow without the correct capitalization.

allyhawkins (approved):
LGTM. I left two small comments, but I don't need to see this again.

PR #25, opened 2024-05-08
added code_deploy

keyrus-rob-jackson:


.github/workflows/run-batch.yml
```diff
@@ -0,0 +1,45 @@
+name: Batch CodeDeploy CI
+
+on:
+  workflow_dispatch:
```
jaclyn-taroni, line 4: I would expect we want this to run on semantically versioned tags:

```suggestion
on:
  push:
    tags: [ v*.*.* ]
  workflow_dispatch:
```

scripts/run_nextflow.sh
```diff
@@ -0,0 +1,2 @@
+unzip -o /opt/nextflow/batch.zip
```
jaclyn-taroni, line 1: Unless there's a reason to exclude?

```suggestion
#!/bin/bash

unzip -o /opt/nextflow/batch.zip
```

.github/workflows/run-batch.yml
```diff
@@ -0,0 +1,45 @@
...
+    environment: "prod"
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::992382809252:role/GithubOpenId
```
jaclyn-taroni, line 23: Can we update the trust policy `sub` to only the `prod` environment (`repo:AlexsLemonade/OpenScPCA-analysis:environment:prod`), then?

jaclyn-taroni: To check my understanding – this is the same role as in `OpenScPCA-analysis` (https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/422/files#diff-dbade1e5797a7633e024a7685a43a7cbd96ae6fefc6314294807ae34cd83712cR25) for convenience, not because there's some limitation on the number of roles that can use this identity provider in their trust policy?

jaclyn-taroni (approved):
👍🏻 

Please feel free to merge without any changes when you are ready. I wanted to jot down my thoughts should we want to make changes in the future.

I did have one question about using the same role for both repos to made sure I understood. 

Thank you!

scripts/run_nextflow.sh
keyrus-rob-jackson: This was no longer needed.  Codedeploy is handing this.

.github/workflows/run-batch.yml
keyrus-rob-jackson: We can use a different role.  This was just to get this working, then we could make adjustments after we verify it is all working.  I will create a different role so we don't need to make that particular change later.

PR #31, opened 2024-05-20
Use new public buckets

### LINKED ISSUE #30: Update simulation output and simulated profile to use current public buckets.
The profile defaults for the workflow currently point to temporary buckets that were used for simulation results and the output of workflows run on simulated results. 

We will need to update these to the correct public buckets: `openscpca-test-data-release-public-access` and `openscpca-test-workflow-results-public-access`

jashapiro:
Closes #30

Updates the buckets used by the simulation module and the `simulated` profile to the new public buckets for openscpca.


jaclyn-taroni (approved):
🚀

PR #37, opened 2024-05-20
add tmux script

### LINKED ISSUE #33: Use background launch of nextflow
Because timeouts for running nextflow may be an issue, we will probably want to launch nextflow into a background process as part of the codedeploy process. 

Ideally, we would use `tmux` or similar to do this, which will allow us to login later to track the workflow in progress.

jashapiro:
Closes #33

This script adds a small wrapper script that launches a new detatched tmux session in which to run the nextflow script. This small PR is a proof of concept, so we can revert it easily if something breaks!

### DIFF
--- a/.github/workflows/run-batch.yml
+++ b/.github/workflows/run-batch.yml
@@ -17,10 +17,10 @@ jobs:
 
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


jaclyn-taroni (approved):
🕯️

jashapiro:
> 🕯️

Success!

PR #38, opened 2024-05-20
Add upload of logs and reports to S3

### LINKED ISSUE #36: Upload logs to S3 post run
When running via a deploy, we will want to upload logs and output summary files (traces and html reports) to S3 so we can look at the results of each run as we go. We will want to modify the run_nextflow script to include those steps.

jashapiro:
Closes #36

This PR adds a nextflow html report to the output of running nextflow, and then uploads that to S3 in the same directory as work directories. I am putting the files in subdirectories by date for now, which may not be what we want at the end (maybe by release tag?), but this is a good place to start, I figure.

I tested the connection from the EC2 instance to the S3 work bucket, and everything there seems to be correct with the required permissions for the upload.

### DIFF
--- a/scripts/run_nextflow.sh
+++ b/scripts/run_nextflow.sh
@@ -1,3 +1,20 @@
 #!/bin/bash
+
+date=$(date "+%Y-%m-%d")
+datetime=$(date "+%Y-%m-%dT%H%M")
+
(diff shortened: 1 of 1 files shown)


jaclyn-taroni (approved):
👍🏻

PR #39, opened 2024-05-21
Pull workflow from GitHub in AWS

### LINKED ISSUE #35: Pull nextflow action from github before running
Currently the running action zips up the entire* repository before submitting that to the AWS node for running. It may be useful both for flexibility and tracking to instead have Nextflow pull a tagged version of the workflow directly and run using that tagged version. We would then zip up only the `scripts` folder for transmission via the aws codedeploy process

jashapiro:
Closes #35

This is the next step in my methodical testing of changes to the nextflow deployment process.

I am now creating a tmux launch script inline in the GHA and letting Nextflow pull the revisions directly from github based on a revision tag. I set the revision tag as a variable the launch script, which should then be picked up in tmux environment and then the `run_nextflow` script. 

This means we now also only zip up the scripts directory, as the remainder of the repo should not be needed.

### DIFF
--- a/.github/workflows/run-batch.yml
+++ b/.github/workflows/run-batch.yml
@@ -4,6 +4,11 @@ on:
   push:
     tags: [v*.*.*]
   workflow_dispatch:
+    inputs:
+      revision:
+        description: "branch or tag to deploy"
+        required: true
+        default: "main"
 
 permissions:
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
🕯️

Issue #32, opened 2024-05-20
Update documentation with entrypoints, profiles, and triggers


jashapiro:
The workflow as it has recently been updated added a few new features in the form of the ability to run specific subsections of the workflow using entrypoints, and profiles to properly use simulated test data.

This issue is to track adding more structured documentation of these features, including how the workflow can be invoked for manual runs, as well as the automated triggers that are implemented as github actions/AWS codedeploy

PR #40, opened 2024-05-21
34 Allow specification of entrypoints in workflow dispatch

### LINKED ISSUE #34: Allow specification of entrypoints in workflow dispatch
When running with workflow dispatch, we probably want to have different entry points that might be triggered for the nextflow workflow (test, simulate, etc). We will add these as inputs to the workflow dispatch trigger and then use them as needed in the action.

jashapiro:
closes #34

Here I am adding a dropdown to the workflow menu to allow running the workflow with various entrypoints. 

These are:
-`test`: runs the minimal test workflow which only tests infrastructure, though we could potentially modify this to run other tests in the future...
- `simulated`: runs the simulation workflow first, then runs the main workflow on the simulated data to generate simulated results
- `full` (don't love this name... maybe "release"?): Runs the full workflow with real data.

Some thoughts on reflection, for which I would like input: 
- Maybe I should split off the actual simulation and only run that on request?
- Alternatively, should the "release" or "full" version run twice: once with the simulated data and once with the real? Part of me things that is the right move, as we would almost always want to keep those in sync.

Note that I will include docs for these options in #32

I will not merge this until I have had a chance to test #39, which is currently held up by GHA being "degraded"

### DIFF
--- a/scripts/run_nextflow.sh
+++ b/scripts/run_nextflow.sh
@@ -3,24 +3,61 @@
 set -u
 
 GITHUB_TAG=${GITHUB_TAG:-main}
+RUN_MODE=${RUN_MODE:-test}
 
+profile="batch"
 date=$(date "+%Y-%m-%d")
 datetime=$(date "+%Y-%m-%dT%H%M")
 
 cd /opt/nextflow
 nextflow pull AlexsLemonade/OpenScPCA-nf -revision $GITHUB_TAG
 
+# test mode runs the test workflow only, then exits
+if [ "$RUN_MODE" == "test" ]; then
+  nextflow run AlexsLemonade/OpenScPCA-nf \
+    -revision $GITHUB_TAG \
+    -entry test \
+    -profile $profile \
+    -entry test \
+    -with-report ${datetime}_test_report.html \
+    -with-trace  ${datetime}_test_trace.txt
+
+  cp .nextflow.log ${datetime}_test.log
+
+  # Copy logs to S3 and delete if successful
+  aws s3 cp . s3://openscpca-nf-data/logs/${RUN_MODE}/${date} \
+    --recursive \
+    --exclude "*" \
+    --include "${datetime}_*" \
+    && rm ${datetime}_*
+  exit 0
+fi
+
+# for simulated mode, run the data simulation pipeline first
+if [ "$RUN_MODE" == "simulated" ]; then
+  nextflow run AlexsLemonade/OpenScPCA-nf \
+    -revision $GITHUB_TAG \
+    -entry simulate \
+    -profile $profile \
+    -with-report ${datetime}_simulate_report.html \
+    -with-trace  ${datetime}_simulate_trace.txt
+
+  cp .nextflow.log ${datetime}_simulate.log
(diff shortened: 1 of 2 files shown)


scripts/run_nextflow.sh
```diff
@@ -3,24 +3,61 @@
...
 cd /opt/nextflow
 nextflow pull AlexsLemonade/OpenScPCA-nf -revision $GITHUB_TAG
 
+# test mode runs the test workflow only, then exits
+if [ "$RUN_MODE" == "test" ]; then
+  nextflow run AlexsLemonade/OpenScPCA-nf \
+    -revision $GITHUB_TAG \
+    -entry test \
+    -profile $profile \
+    -entry test \
```
jaclyn-taroni, line 21: I assume?
```suggestion
```

scripts/run_nextflow.sh
```diff
@@ -3,24 +3,61 @@
...
+  # Copy logs to S3 and delete if successful
+  aws s3 cp . s3://openscpca-nf-data/logs/${RUN_MODE}/${date} \
+    --recursive \
+    --exclude "*" \
+    --include "${datetime}_*" \
+    && rm ${datetime}_*
+  exit 0
+fi
+
+# for simulated mode, run the data simulation pipeline first
```
jaclyn-taroni, line 36: So if you select simulated, the full will also run? I'm not sure I would expect that. (I understand docs are coming!)

jaclyn-taroni: I went back and re-read your comments. I'm leaving this here so you have my genuine reaction. 😄 

I agree that we probably want to keep these in sync; the use of `simulated` would make me think it's just simulated.

I wonder if it should be something like: `simulated` (simulated only), `release` (ScPCA data – I also thought `scpca` might work), and `full` (simulated and ScPCA data, where you'd set this as the default.)

jaclyn-taroni (commented):
Returning my input! And one bug, I think.

scripts/run_nextflow.sh
jashapiro: Okay, I think I will go with `simulated`, `scpca`, and `full`, but in a change from the current setup, the _simulation_ step will only run with `full`, which is what will be triggered with a workflow release. 

Which is to say: 
-`simulated`: run the main workflow with simulated data
-`scpca`: run the main workflow with real data
-`full`: simulate data based on real data, then run the main workflow once with the simulated data and once with the real data
-`test`: same as before

Should I also have a `simulate-only`? Or am I wrong and the `simulated` workflow should also run the simulation component?

jaclyn-taroni: > -`simulated`: run the main workflow with simulated data
> -`scpca`: run the main workflow with real data
> -`full`: simulate data based on real data, then run the main workflow once with the simulated data and once with the real data
> -`test`: same as before

I think this is right

jashapiro:
Now that I got things working again, I have updated with new options as we discussed in the comments. Should be ready for review!

jaclyn-taroni (approved):
Returning a bug fix. LGTM otherwise 👍🏻

PR #41, opened 2024-05-21
Fix deploy?

jashapiro:
Apparently I broke something. The deploy seems to be failing at the `beforeInstall` phase, even though I did not do anything to that. But there is nothing really happening in that script, so I just took that hook out of the appspec, and lets see if that fixes things.


jaclyn-taroni (approved):
🕯️

PR #42, opened 2024-05-21
Add appspec to zip file

jashapiro:
I wonder if this was failing for not including the appspec in the zip file, so now I am doing that too.


jaclyn-taroni (approved):
That makes some sense to me

PR #53, opened 2024-05-29
Use resourceLimits and add test runs

jashapiro:
This started with just changing to using the new `process.resourceLimits` to replace our functions for checking memory and cpus against the maximum allowances. 

One thing that was a bit non-obvious was how to handle the `mem_max` label: I ended up just picking a memory amount larger than any max value we are likely to use in the directive, but this is automatically scaled down to whatever is set in the `process.resourceLimits` setting. 

The one disadvantage of this setup is that it is a bit harder to change limits on the fly; you can't just set a `param` and have that propagate in as we had before, but I don't expect that to be a common use case.

After I added the resourceLimits changes, I thought this was a good chance to add some more testing, so I included a new GitHub action for stub runs of the workflow.

### DIFF
--- a/.github/workflows/nextflow-stub-check.yml
+++ b/.github/workflows/nextflow-stub-check.yml
@@ -18,12 +18,12 @@ jobs:
       - name: Check main workflow
         uses: docker://nextflow/nextflow:24.04.2
         with:
-          args: nextflow run main.nf -stub -profile stub -ansi-log false -log stub-run.log
+          args: nextflow -log stub-run.log run main.nf -stub -profile stub -ansi-log false
 
       - name: Check simulate workflow
         uses: docker://nextflow/nextflow:24.04.2
         with:
-          args: nextflow run main.nf -stub -profile stub -ansi-log false -entry simulate -log simulate-run.log
+          args: nextflow -log simulate-run.log run main.nf -stub -profile stub -ansi-log false -entry simulate
 
       - name: Join log files
         if: ${{ !cancelled() }}
--- a/.github/workflows/nextflow-stub-check.yml
+++ b/.github/workflows/nextflow-stub-check.yml
@@ -11,16 +11,19 @@ on:
 jobs:
   nf-stub-check:
     runs-on: ubuntu-latest
-    container: nextflow/nextflow:24.04.2
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
       - name: Check main workflow
-        run: nextflow run main.nf -stub -profile stub -ansi-log false -log stub-run.log
+        uses: docker://nextflow/nextflow:24.04.2
+        with:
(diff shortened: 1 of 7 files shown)


.github/workflows/nextflow-stub-check.yml
```diff
@@ -0,0 +1,37 @@
+name: Check nextflow stub
+
+on:
+  push:
+    branches:
+      - jashapiro/resourceLimits
```
allyhawkins, line 6: Just a note that we should remove this before merging?

config/process_base.config
```diff
@@ -12,47 +12,35 @@ process {
...
   withLabel: mem_32 {
-    memory = {check_memory(32.GB * task.attempt, params.max_memory)}
+    memory = { 32.GB * task.attempt }
   }
-  withLabel: mem_max { // max memory for tasks that have failed more than twice for OOM
-    memory = {(task.attempt > 2  && task.exitStatus in 137..140) ? params.max_memory : check_memory(64.GB * task.attempt, params.max_memory)}
+  withLabel: mem_max {
+    // max memory for tasks that have failed more than twice for OOM
+    // set to 2.TB, but will be reduced by process.resourceLimits
+    memory = {(task.attempt > 2  && task.exitStatus in 137..140) ? 2.TB : 64.GB * task.attempt }
```
allyhawkins, line 29: Just noting that I think it's fine to keep this set at 2 TB. Although if you are going to have to actually change the config files to change the resource limits anyways, why not just set this to whatever the resource limit is?

nextflow.config
```diff
@@ -17,24 +17,20 @@ params {
...
+      executor = 'local'
+      resourceLimits = [ cpus: 4, memory: 16.GB ]
+    }
     docker.enabled = true
     docker.userEmulation = true
-    params {
-      max_cpus = 8
-      max_memory = 32.GB
-    }
+    params.max_memory = 16.GB
```
allyhawkins, line 33: I don't think you are actually using this anymore?

allyhawkins (commented):
I agree that it's a bit annoying to not be able to set the options at the command line, but if we come across running a lot of processes that use > 512 GB then we can revisit this. I think that's highly unlikely though. I just had a few small comments.

config/process_base.config
jashapiro: Yeah, I figured it was easier to have only one place to change/update. There is a small risk we will forget that there is an implicit 2TB max, but if we are running into that, we probably have other problems as well!

.github/workflows/nextflow-stub-check.yml
jashapiro: Thanks for catching this & the lingering `params.max_memory`

Should be ready to go now.

allyhawkins (approved):
LGTM

PR #54, opened 2024-05-30
49 Add filtering to merge module to exclude known failures

### LINKED ISSUE #49: Add filtering to merge module to exclude known failures
In #45  and from other experience, we know that SCPCP000008 is too large to be successfully exported as AnnData. We should add a quick filterer to the merge module to exclude specific projects, or perhaps just libraries that have too many samples for futureproofing.

jashapiro:
Closes #49

To reduce computation for the merge workflow a bit when we expect it to fail, I am here adding a parameter for the maximum number of libraries to merge.

I guessed on the max to be between the failure and success, but as it is a parameter it should be easy to test higher numbers if we get libraries near the boundary.

### DIFF
--- a/modules/merge-sce/main.nf
+++ b/modules/merge-sce/main.nf
@@ -131,13 +131,23 @@ workflow merge_sce {
         def library_ids = processed_files.collect{it.name.replace('_processed.rds', '')}
         return [project_id, library_ids, processed_files]
       }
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
This looks good to me. I think 75 is a reasonable guess for now.

Issue #29, opened 2024-05-10
Discussion: Inputs and outputs for workflow modules


jashapiro:
Currently, the inputs for OpenScPCA-nf modules are channels that contain the following tuple:
`[project_id, project_dir]`

This makes for a pretty nice compact and universal input, and for modules that depend on nothing but "raw" data it is is pretty easy to work with. It is therefore currently the module workflow's responsibility to filter the inputs down to the portion of the data that is needed for each analysis. 

However, that structure may not be workable in the case that one module depends on the output of another module. We have not yet described standards for what a module workflow might `emit` for consumption by other module workflows, but it seems unlikely that the output would be of the form `[project_id, project_dir]`. 

There are a couple reasons for this: 
- For efficiency, workflows are likely to split input projects into multiple streams, and joining them back into a single output could be difficult
	- For example, the `simulate-sce` workflow largely performs its simulations on a single sample at a time. While it would be possible to rejoin the results into a single directory, this would mean we would need to have a process that downloads the data for all samples within each project and then emits the folder with all of them in it. This would result in a lot of copying without much benefit.
- The contents of an output directory may not be as predictable as inputs, and it may be better to have more fully specified input channels which contain more information about the kind of data that is needed for each workflow. 

## A motivating example

I already mentioned the `simulate-sce` module workflow, and it was the main reason I started thinking about this problem in the first place (though as I write this I may have a simpler/better solution, discussed below). The `simulate-sce` module workflow does not currently `emit` anything, but we might want it to, in order to use the results from that workflow in downstream workflows. For example, the `merge-sce` workflow could be used to create simulated merged data. 

However, the `merge-sce` workflow currently expects `[project_id, project_dir]` as input, and as discussed, this might be a pain to produce from the simulation workflow, which works on a sample-by-sample basis. 

An easier output might be something of the form `[project_id, sample_id, sample_dir]`, but then the consuming workflow (`merge-sce` in this case) might have to perform some form of merging on the input channel if it wants things organized by project. This isn't necessarily terrible to do: we could do that with something like `groupTuple()` function.

[...]

PR #55, opened 2024-05-30
Use samples as principal workflow elements

jashapiro:
Here I am changing the inputs to the merge workflow to take a channel of samples rather than a channel of projects.

This reflects the discussion in #29 about what the "standard" input to a module should be. In this case it doesn't matter much, as the sample channel is condensed down to the former project channel, but this presents an example of how this might be done in the future. 

I also added/modified a few of the `Util` functions to return lists of tuples that can be converted to channels with the `fromList()` factory method, rather than having to build the channels from the paths as before.

I left the input for the simulation script alone, partly just to minimize changes, and partly to show that either might be possible as we proceed to add modules in the future.

### DIFF
--- a/lib/Utils.groovy
+++ b/lib/Utils.groovy
@@ -32,17 +32,23 @@ class Utils {
     return projects
   }
 
-  static def getProjectPaths(release_path){
-    release_path = Nextflow.file(release_path, type: 'dir')
-    def projects = getProjects(release_path)
-    return projects.collect{release_path / it}
+  static def getProjectTuples(release_path){
+    // create a list of tuples of [project_id, project_path]
+    def project_paths = Nextflow.files(release_path / "SCPCP*", type: 'dir')
+    return project_paths.collect{new Tuple(it.name, it)}
   }
 
-  static def getProjectFiles(Map args, project_dir){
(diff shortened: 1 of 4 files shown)


allyhawkins (approved):
This LGTM, but are we planning to only ever run this for an entire project? If not, we probably want to add a param for samples and functionality to run for only a subset of samples from a project?

jashapiro:
> This LGTM, but are we planning to only ever run this for an entire project? If not, we probably want to add a param for samples and functionality to run for only a subset of samples from a project?

I'm not yet sure. Some modules may want to filter to only some samples, but I think that filtering should happen at the module level, not the top level, as different modules may require different filtering.

jashapiro:
> > This LGTM, but are we planning to only ever run this for an entire project? If not, we probably want to add a param for samples and functionality to run for only a subset of samples from a project?
> 
> I'm not yet sure. Some modules may want to filter to only some samples, but I think that filtering should happen at the module level, not the top level, as different modules may require different filtering.

I'm actually not sure if the project filtering should happen at the top level either. The only reason it is there now is for the `stub` run, which may be a bad idea as we get to modules that only work on specific projects. But it is easy enough to remove/modify if needed.

PR #57, opened 2024-06-04
Use staging buckets by default

### LINKED ISSUE #44: Add testing/staging output buckets
Currently we have only the "production" output buckets specified in the Nextflow workflow, but this is probably not the ideal way to have the defaults set up, as any errors in the workflow could result in corrupted data in the buckets used by other tools/the `scpca-analysis` scripts and modules .

We should probably change the defaults to some kind of staging location, and copy the results over in a separate step, or with a rerun and the final output bucket specified.  This may take making a few more buckets for these purposes.

jashapiro:
Closes #44 

Here I am adding use of the staging buckets that were created in https://github.com/AlexsLemonade/OpenScPCA-infra/pull/33 to the workflow.

The idea is that most runs initiated through workflow triggers will use the staging buckets, with the production buckets reserved for tagged releases. I am not currently enforcing that: setting a run to `prod` is still possible from the GHA, though I wonder if I should instead make it _only_ run on version tags.

I moved the setting of the simulated data bucket out to a more global param to make it more visible and consistent, but note that I left Docker image definitions with the code that they are used for.

I also updated the readme to cover these options and to describe which buckets are being used for the action outputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The workflow is currently set up to run best via AWS batch, but some testing may
 You will need to have appropriate AWS credentials set up to run the workflow on AWS and access the data files.
 In general, you must have `workload` access in an OpenScPCA AWS account to run the workflow.
 
-### Running the workflow from GitHub Actions
+### Running the workflow using GitHub Actions
 
 The most common way to run the workflow will be to run the GitHub Action (GHA) responsible for running the workflow.
 The GHA is run automatically when a new release tag is created or by manually triggering the workflow.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -26,14 +26,26 @@ The GHA workflow will run automatically when a new release tag is created, which
...
+
 The run modes available are:
 
 - `test`: runs only a simple test workflow to check configuration
 - `simulated`: runs the workflow using simulated data
 - `scpca`: runs the workflow using the current ScPCA data release
 - `full`: simulates data based on the current ScPCA data release, then runs the workflow using the simulated data and current ScPCA data release (this is same as the behavior of the automatic release workflow)
 
+By default, the output mode will be set to `staging`, so all outputs will be saved to S3 buckets that are not shared with users and can not overwrite current production data.
+With the `prod` output mode, results will be accessible visible to users.
```
allyhawkins, line 39:
```suggestion
With the `prod` output mode, results will be accessible to users.
```

allyhawkins: Maybe add a note that prod should only be used for releases?

allyhawkins (approved):
LGTM, I would just add a small note that the `prod` profile is meant to be used for releases.

README.md
allyhawkins: Looks good!

PR #58, opened 2024-06-05
Add monitoring of OpenScPCA-nf nextflow jobs

### LINKED ISSUE #46: Add notification of nextflow workflow failure
Right now the only way to know that a failure has occurred after running the workflow via the CodeDepoly is to look at the log files that were uploaded to S3, or on the Nextflow-workloads server.

It would be nice to have a more direct notification of failure. I am sure there are some recommended ways to do this, but I do not currently know the best practice.

jashapiro:
I added a couple new ways to monitor nextflow runs.

1. Posts to Slack when there are errors during nextflow runs. This is implemented through a slack webhook

2. Seqera tower (formerly Nextflow tower) monitoring. This is implemented with a Workspace, so individuals can log in and monitor from their own accounts if they have one and are part of the CCDL org, or through the scpca@ccdatalab.org account (logging in via Google). An example of a run is at https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/608P3fD4SFXsQ6

To do this, I also implemented getting secrets via 1Password/AWS secrets manager, which seems to be working.

I updated the readmes as well, separating out internal instructions, including monitoring info, from the more general readme for people who might want to run on their own infrastructure. 

Closes #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,8 @@ The run modes available are:
 - `full`: simulates data based on the current ScPCA data release, then runs the workflow using the simulated data and current ScPCA data release (this is same as the behavior of the automatic release workflow)
 
 By default, the output mode will be set to `staging`, so all outputs will be saved to S3 buckets that are not shared with users and can not overwrite current production data.
-With the `prod` output mode, results will be accessible visible to users.
+With the `prod` output mode, results will be accessible to users.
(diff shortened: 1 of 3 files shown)


jaclyn-taroni (approved):
👍🏻

PR #67, opened 2024-07-10
Always output simulated data to `test/` prefix

### LINKED ISSUE #66: Simulate script has wrong publish location
The simulate script is supposed to publish to `s3://openscpca-test-data-release-public-access/test/`, but it seems like the last run published to `s3://openscpca-test-data-release-public-access/2024-07-08/`

I will move the files to the correct location manually, but also investigate why this happened.

jashapiro:
Closes #66

The issue in #66 was that I was using a variable rather than hard coding the `test/` prefix for simulated data output. This should fix that. I had considered adding another parameter, but that did not seem worth it.

I also ran a shell linter on the bash script as a test, and this includes a number of suggestions from that, with no expected changes to behavior (mostly just quoting things)


allyhawkins (approved):
LGTM

PR #72, opened 2024-07-16
Add doublet detection module

### LINKED ISSUE #68: Add a doublet detection module
This issue tracks adding a `doublet-detection` module (see original `OpenScPCA-analysis` module [here](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/doublet-detection)) which runs `scDblFinder` on SCEs to the `OpenScPCA-nf` workflow. This module should take a single processed SCE as input, and export a TSV with doublet predictions. Files are processed and results are organized by projects and samples, following the same directory structure as how the OpenScPCA data is organized.

This issue will encompass three tasks, which might be combined in PRs but may also be separate:
- [ ] Port the doublet script
- [ ] Add the doublet module
- [ ] Call the doublet workflow from `main.nf`

sjspielman:
Closes #68 

This PR adds the doublet detection module to the workflow. Currently it is very much not working (and note I temporarily have a hardcoded local path in there for results), but it might be close to working! Either way, now is a good time for me to get some feedback before I go too far with this maybe-not-ideal apparoach.

I decided to begin implementing this by borrowing some code from the `merge_sce` module which I can use to grab all library ids for a given sample. I need the library ids in order to `touch` the stub output. Currently, it's errorring b/c it doesn't like my for loop `library_ids` variable, but I am quite sure something else is wrong with my `output_dir` variable (which maybe I shouldn't be trying to use anyways, and instead update the doublets script to take an output TSV file path rather than a directory where that TSV file lives). Note that `merge_sce()` is also temporarily commented out.

### DIFF
--- a/modules/doublet-detection/README.md
+++ b/modules/doublet-detection/README.md
@@ -1,0 +1,1 @@
+This module performs doublet detection on processed SCE objects with [`scDblFinder`](https://bioconductor.org/packages/release/bioc/html/scDblFinder.html).
--- a/modules/doublet-detection/README.md
+++ b/modules/doublet-detection/README.md
@@ -1,1 +1,5 @@
 This module performs doublet detection on processed SCE objects with [`scDblFinder`](https://bioconductor.org/packages/release/bioc/html/scDblFinder.html).
+
+Scripts are derived from the the `doublet-detection` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
+
+Permalink to the version used: <incoming>
--- a/modules/doublet-detection/README.md
+++ b/modules/doublet-detection/README.md
@@ -2,4 +2,4 @@ This module performs doublet detection on processed SCE objects with [`scDblFind
 
 Scripts are derived from the the `doublet-detection` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
 
-Permalink to the version used: <incoming>
+Permalink to the version used: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/cdc71e4ac0ca4fcaf6ac225002c928453913f734/analyses/doublet-detection/scripts/01a_run-scdblfinder.R>
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -34,7 +34,7 @@ process run_scdblfinder {
   stub:
     output_files = library_files
       .collect{
-        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv"))
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
       }
     """
     for file in ${library_files}; do
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -19,9 +19,8 @@ process run_scdblfinder {
   script:
     output_files = library_files
       .collect{
-        file(it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")).getName()
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
       }
-    //print(output_files)
 
     """
     for file in ${library_files}; do
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -11,21 +11,35 @@ process run_scdblfinder {
   input:
     tuple val(sample_id),
           val(project_id),
-          path(processed_file),
-          val(output_file)
+          path(library_files)
   output:
-     tuple val(sample_id), path(output_file)
+    tuple val(sample_id),
+          val(project_id),
+          path(output_files)
   script:
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
+
     """
-    ./run_scdblfinder.R \
-      --input_sce_file "${processed_file}" \
-      --output_file "${output_file}" \
-      --random_seed 2024 \
-      --cores ${task.cpus}
+    for i in "${!library_files[@]}"; do
+      ./run_scdblfinder.R \
+        --input_sce_file "${library_files[\$i]}" \
+        --output_file "${output_files[\$i]}" \
+        --random_seed 2024 \
+        --cores ${task.cpus}
+    done
     """
   stub:
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
     """
-    touch ${output_file}
+    for output_file in ${output_files}; do
+      touch \${output_file}
+    done
     """
 }
 
@@ -36,18 +50,11 @@ workflow detect_doublets {
     sample_ch  // [sample_id, project_id, sample_path]
   main:
 
+    // create [sample_id, project_id, [list, of, processed, files]]
     libraries_ch = sample_ch
-      // create [sample_id, project_id, [list, of, processed, files]]
       .map{sample_id, project_id, sample_path ->
         def library_files = Utils.getLibraryFiles(sample_path, format: "sce", process_level: "processed")
         return [sample_id, project_id, library_files]
-      }
-      // create [sample_id, project_id, processed file], for all files
-      .transpose()
-      // create [sample_id, project_id, processed file, output file]
-      .map{sample_id, project_id, library_file ->
-        def output_file = library_file.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
-        return [sample_id, project_id, library_file, output_file]
       }
 
     // detect doublets
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -23,10 +23,10 @@ process run_scdblfinder {
       }
 
     """
-    for i in "${!library_files[@]}"; do
+    for file in ${library_files}; do
       ./run_scdblfinder.R \
-        --input_sce_file "${library_files[\$i]}" \
-        --output_file "${output_files[\$i]}" \
+        --input_sce_file \$file \
+        --output_file \$(basename \${file%.rds}_scdblfinder.tsv) \
         --random_seed 2024 \
         --cores ${task.cpus}
     done
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -21,7 +21,6 @@ process run_scdblfinder {
       .collect{
         it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
       }
-
     """
     for file in ${library_files}; do
       run_scdblfinder.R \
@@ -31,6 +30,7 @@ process run_scdblfinder {
         --cores ${task.cpus}
     done
     """
+
   stub:
     output_files = library_files
       .collect{
@@ -49,7 +49,6 @@ workflow detect_doublets {
   take:
     sample_ch  // [sample_id, project_id, sample_path]
   main:
-
     // create [sample_id, project_id, [list, of, processed, files]]
     libraries_ch = sample_ch
       .map{sample_id, project_id, sample_path ->
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -37,9 +37,7 @@ process run_scdblfinder {
         it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
       }
     """
-    for output_file in ${output_files}; do
-      touch \${output_file}
-    done
+    touch  ${output_files}
     """
 }
 
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -7,7 +7,7 @@ params.doublet_detection_container = 'public.ecr.aws/openscpca/doublet-detection
 process run_scdblfinder {
   container params.doublet_detection_container
   tag "${sample_id}"
-  publishDir "/Users/sjspielman/ALSF/doublet-detection/${project_id}", mode: 'copy'
+  publishDir "${params.results_bucket}/${params.release_prefix}/doublet-detection/${project_id}/${sample_id}", mode: 'copy'
   input:
     tuple val(sample_id),
           val(project_id),
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -11,25 +11,21 @@ process run_scdblfinder {
   input:
     tuple val(sample_id),
           val(project_id),
-          path(sample_path),
-          val(library_list)
+          path(processed_file),
+          val(output_file)
   output:
-    val(sample_id)
+     tuple val(sample_id), path(output_file)
   script:
-    output_dir=file("${sample_id}", type: 'dir')
     """
-    for library_id in ${library_list}; do
-      Rscript run_scdblfinder.R \
-        --input_sce_file "${sample_path}/${library_id}_processed.rds" \
-        --results_dir ${output_dir}
-    done
+    ./run_scdblfinder.R \
+      --input_sce_file "${processed_file}" \
+      --output_file "${output_file}" \
+      --random_seed 2024 \
+      --cores ${task.cpus}
     """
   stub:
-    output_dir=file("${sample_id}", type: 'dir')
     """
-    for library_id in ${library_list}; do
-      touch ${output_dir}/${library_id}_scdblfinder.tsv
-    done
+    touch ${output_file}
     """
 }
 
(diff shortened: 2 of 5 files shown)


nextflow.config
```diff
@@ -9,6 +9,7 @@ manifest {
   nextflowVersion = '>=24.04.0'
 }
 
+nextflow.enable.dsl = 2
```
jashapiro, line 12: I guess it doesn't hurt to have this, but it is the default with the required nextflow version listed above, so it should not be required.

modules/doublet-detection/README.md
```diff
@@ -0,0 +1 @@
+This module performs doublet detection on processed SCE objects with [`scDblFinder`](https://bioconductor.org/packages/release/bioc/html/scDblFinder.html).
```
jashapiro, line 1: Please include a permalink to the version of the analysis repo that this module is derived from.

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    sample_ch  // [sample_id, project_id, sample_path]
+  main:
+
+    // create [sample_id, project_id, sample_path, [list, of, library, ids, in, sample, path]]
+    libraries_ch = sample_ch
+      .map{ sample_id, project_id, sample_path ->
+        def processed_files = Utils.getLibraryFiles(sample_path, format: "sce", process_level: "processed")
+        def library_ids = processed_files.collect{it.name.replace('_processed.rds', '')}
+        return [sample_id, project_id, sample_path, library_ids]
+      }
```
jashapiro, line 49: It doesn't look like you are using the library ids at all, so you should be able to simplify this:

```suggestion
      .map{ sample_id, project_id, sample_path ->
        def processed_files = Utils.getLibraryFiles(sample_path, format: "sce", process_level: "processed")
        return [sample_id, project_id, processed_files]
      }
```

Note that `getLibraryFiles` returns a files object (an array of files), so you shouldn't need the `sample_path` at this point.

The idea here is that we will not pass the whole sample directory, but _only_ the files we need. This will save a lot of bandwidth.

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,53 @@
...
+
+process run_scdblfinder {
+  container params.doublet_detection_container
+  tag "${sample_id}"
+  publishDir "/Users/sjspielman/ALSF/doublet-detection/${project_id}", mode: 'copy'
+  input:
+    tuple val(sample_id),
+          val(project_id),
+          path(sample_path),
+          val(library_list)
```
jashapiro, line 15:
```suggestion
    tuple val(sample_id),
          val(project_id),
          path(library_files)
```

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,53 @@
...
+  output:
+    val(sample_id)
```
jashapiro, line 17: Output should be a tuple of 
```suggestion
    tuple val(sample_id), path(output_files)
```

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,53 @@
...
+  script:
+    output_dir=file("${sample_id}", type: 'dir')
```
jashapiro, line 19: Here you will probably need to write something that creates the output file names that your script generates. 

Based on how your script currently works, it might be something like:
```
output_files = library_files.collect{
  file($sample_id, type: 'dir') / it.baseName.replaceAll(/(?i).rds$/, '_scdblfinder.tsv')
}
```

We could use just a normal `.replace()`, but that does not allow a regex to match the end of the string for better specificity and case insensitivity with `(?i)`, which seem like nice additions. (I just learned that and I will probably want to apply it elsewhere in this repo)

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,53 @@
...
+    """
+    for library_id in ${library_list}; do
+      Rscript run_scdblfinder.R \
+        --input_sce_file "${sample_path}/${library_id}_processed.rds" \
+        --results_dir ${output_dir}
+    done
```
jashapiro, line 25: One of the goals with nextflow is to try to avoid dealing with paths as much as possible within the processes. If a file comes in as a file, we can just use it directly.

We need to escape `$` to get them to work as bash variables: by default they are Nextflow variables.

If you use my suggestion before, we can just put the output file into the current directory, which will make it easy for nextflow to pick it up.

Finally, you need to call your script directly (which means making it executable if it not already) 

```suggestion
    for infile in ${library_files}; do
      run_scdblfinder.R \
        --input_sce_file "\$infile" \
        --results_dir $sample_id
    done
```

modules/doublet-detection/resources/usr/bin/run_scdblfinder.R
```diff
@@ -0,0 +1,175 @@
...
+}
+
+output_tsv_file <- file.path(
+  opts$results_dir,
+  stringr::str_replace(
+    basename(opts$input_sce_file),
+    ".rds",
+    "_scdblfinder.tsv"
+  )
+)
```
jashapiro, line 127: Okay, you convinced me. You should change this script to take an output file name, rather than a directory. Then your wrapper shell scripts can change this more easily and the whole thing becomes more flexible.

modules/doublet-detection/main.nf
jashapiro: You will also want to set a seed here, and give the number of cores using `${task.cpus}`

modules/doublet-detection/main.nf
jashapiro: Note that you can use the same code as above in the `stub` for the list of files to touch

jashapiro (commented):
Yes, it was a good time to reach out! I think I agree that having the script take a file path for the output would be a good change, as it makes everything later more flexible. In general, that is my preferred mode unless a script is outputting many files.

But the workflow can work either way. The main thing is that for the input you want to avoid passing in the whole sample directory, as that contains _all_ versions of the SCE/Anndata objects, so it is quite large. The idea of the `Utils.getLibraryFiles()` function that you found is to avoid that completely by allowing you to get an array of paths/files to pass along to a process (it is always an array for consistency, even if just one file is being passed). That will mean that only the files you need are copied to the worker node. (The merge script needs the `library_id`s too for labelling, but you should not need that, and just the files is sufficient).

Hopefully the suggestions I made make sense... let me know if you have further questions. We can also set a time to pair on this, which may be the most efficient.

nextflow.config
sjspielman: Nextflow specifically complained to me about this, is why I added it.

jashapiro: It should have complained that you weren't using `v24.04`, as specified a couple lines earlier... I think that version may be when the default was officially changed over.

modules/doublet-detection/main.nf
sjspielman: FYI turns out `baseName` also eats up the extension! https://github.com/nextflow-io/nextflow/blob/e562ce061c0133ab2d16d8601127d86e1aba8d55/modules/nf-commons/src/main/nextflow/extension/FilesEx.groovy#L320-L349

jashapiro: oh, maybe better to stick with `it.name` then and do the replacement more conservatively.

sjspielman:
Stub is happy, so I'm marking this actually ready for review!

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env nextflow
+
+// Workflow to detect doublets in a SingleCellExperiment object using scDblFinder
+
+params.doublet_detection_container = 'public.ecr.aws/openscpca/doublet-detection:latest'
```
jashapiro, line 5: Is there another tagged version we can use? The answer may be no for now, but we will want to have tagged versions where we can.

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,55 @@
...
+  main:
+
+    libraries_ch = sample_ch
+      // create [sample_id, project_id, [list, of, processed, files]]
+      .map{sample_id, project_id, sample_path ->
+        def library_files = Utils.getLibraryFiles(sample_path, format: "sce", process_level: "processed")
+        return [sample_id, project_id, library_files]
+      }
+      // create [sample_id, project_id, processed file], for all files
+      .transpose()
```
jashapiro, line 46: I don't think we want to do this, but rather handle the input as a possible array of files. The issue is that this creates duplicate sample_ids, so it makes merging later much harder. I would rather the process ran a small loop for the cases where there is more than one library file

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,55 @@
...
+      // create [sample_id, project_id, processed file, output file]
+      .map{sample_id, project_id, library_file ->
+        def output_file = library_file.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
```
jashapiro, line 49: I think the output files are better defined within the process itself, rather than passed into it. Doing so keeps the `scdblfinder` name directly linked to the action. 

It does make for some duplicated code as you need versions for both the script and stub, but I still think that is preferable.

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,55 @@
...
+
+process run_scdblfinder {
+  container params.doublet_detection_container
+  tag "${sample_id}"
+  publishDir "${params.results_bucket}/${params.release_prefix}/doublet-detection/${project_id}/${sample_id}", mode: 'copy'
+  input:
+    tuple val(sample_id),
+          val(project_id),
+          path(processed_file),
+          val(output_file)
```
jashapiro, line 15: output file probably shouldn't be an input value for the process, but should be computed within.

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,55 @@
...
+  output:
+     tuple val(sample_id), path(output_file)
```
jashapiro, line 17: I know I said just this, but we should probably keep the project id in here just in case. 

```suggestion
     tuple val(sample_id), val(project_id), path(output_file)
```

jashapiro (commented):
While this works, I would still prefer the process to act on the level of samples, not individual files. While you could add a step to merge the samples back together, that actually creates some additional blocking within the nextflow pipeline (we have to wait for all of the channel to finish to know whether we have more to merge). In general I have found that minimizing channel transformations is preferable when possible.

modules/doublet-detection/main.nf
sjspielman: no for now, so I'll open a separate issue to keep track of adding this in

modules/doublet-detection/main.nf
sjspielman: while working to get this back to a for loop (question: would you prefer a branch for multiplexed samples and only loop there?), I encountered a syntax error that I simply could not find...can you?

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
+
+    """
+    for i in "${!library_files[@]}"; do
+      ./run_scdblfinder.R \
+        --input_sce_file "${library_files[\$i]}" \
+        --output_file "${output_files[\$i]}" \
```
jashapiro, line 29: Not worth the effort to try to sync the arrays... I'd just use what we did in the other bash script
```suggestion
    for file in ${library_files}; do
      ./run_scdblfinder.R \
        --input_sce_file \$file \
        --output_file \$(basename \${file%.rds}_scdblfinder.tsv) \
```

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    """
+  stub:
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
+    """
+    for output_file in ${output_files}; do
+      touch \${output_file}
+    done
```
jashapiro, line 42: `touch` can handle multiple files 
```suggestion
    touch  ${output_files}
```

sjspielman:
Any any idea why it's (maybe?) not staging the script...?
https://github.com/AlexsLemonade/OpenScPCA-nf/actions/runs/9978872937/job/27576707385?pr=72#step:4:53

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,60 @@
...
+        --cores ${task.cpus}
+    done
+    """
+  stub:
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
+    """
+    touch  ${output_files}
```
jashapiro, line 40: > Any any idea why it's (maybe?) not staging the script...? https://github.com/AlexsLemonade/OpenScPCA-nf/actions/runs/9978872937/job/27576707385?pr=72#step:4:53

The script is a red herring. It is failing with the stub and when that happens it falls back to the main script, which won't work in the stub environment because docker is required for the file structure we use. 

The issue is that the `.collect` isn't doing quite what we want in that it is giving us back a list of strings, not file paths. Which is kind of fine, because we don't really want the full paths at this stage, but I think it means that nextflow isn't deparsing the array the way it would if it were an array of paths. So we may have to do that manually. 
```suggestion
    touch  ${output_files.join(' ')}
```

I tested this locally, and it does work for the stub. To test what I thought would happen with the main loop I also tried this closer parallel, which also worked:

```suggestion
    for file in ${library_files}; do
      touch \$(basename \${file%.rds}_scdblfinder.tsv)
    done
```

I also want to note how I did this debugging: after running locally, I opened up `.nextflow.log` and then went to the first failure, which included a path to the relevant work directory. In there I looked at `command.sh` to see what the script actually had in it. In this case it had `touch [SCPCP00000_processed_scdblfinder.tsv]` which is what told me there was a parsing problem with the list of strings being converted the same way that lists of paths usually are.

sjspielman: possibly got the strings to be files in https://github.com/AlexsLemonade/OpenScPCA-nf/pull/72/commits/c8f4c7046f47ce4ccc5cb79858d800c7ca2f51fb?

modules/doublet-detection/main.nf
```diff
@@ -34,10 +35,10 @@ process run_scdblfinder {
   stub:
     output_files = library_files
       .collect{
-        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+        file(it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")).getName()
```
jashapiro, line 38: You are just adding work now... Turning it back into a `file` object may add paths you don't want, and I don't think you need. And then you are just stripping them back off with `.getName()`, which was already done with `it.name`

sjspielman: I'm very prepared to accept that I'm adding work, but also this means I don't think I've understood your comment - 

> isn't doing quite what we want in that it is giving us back a list of strings, not file paths. 

I read this as, "we want it to give us back file paths". No?

jashapiro: Maybe, but it doesn't really matter as long as we know what we have and treat it appropriately. But I am saying now that `file(string).getName() == string`. So you aren't doing anything with the added code. 

In some sense: 
file(string) == Path("$pwd/string")

jashapiro: well, that isn't quite accurate... `file(path.name).getName() == path.name`

(Also `.getName() == .name`; any time you see a `get` function you can use just the attribute of what is being get)

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    """
+}
+
+
+
+workflow detect_doublets {
+  take:
+    sample_ch  // [sample_id, project_id, sample_path]
+  main:
+
```
jashapiro, line 52: Remove extra line
```suggestion
```

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,62 @@
...
+  output:
+    tuple val(sample_id),
+          val(project_id),
+          path(output_files)
+  script:
+    output_files = library_files
+      .collect{
+        it.name.replaceAll(/(?i).rds$/, "_scdblfinder.tsv")
+      }
+
```
jashapiro, line 24: remove blank line
```suggestion
```

modules/doublet-detection/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    """
+    for file in ${library_files}; do
+      run_scdblfinder.R \
+        --input_sce_file \$file \
+        --output_file \$(basename \${file%.rds}_scdblfinder.tsv) \
+        --random_seed 2024 \
+        --cores ${task.cpus}
+    done
+    """
+  stub:
```
jashapiro, line 34: Blank line before stub
```suggestion

  stub:
```

modules/doublet-detection/resources/usr/bin/run_scdblfinder.R
```diff
@@ -0,0 +1,167 @@
...
+
+  run_scdblfinder(
+    sce,
+    cores = opts$cores,
+    # only used if there are multiple samples, e.g. if this is processing a merged object
+    sample_var = opts$sample_var, # will be NULL if not provided as input argument
+    random_seed = opts$random_seed,
+    columns_to_keep = keep_columns
+  ) |>
+    readr::write_tsv(opts$output_file)
```
jashapiro, line 166: This is a minor suggestion, mostly because I don't love piping directly to output, especially when the using a custom function with a potentially complex output.  But making this change here would require a change in `scpca-analysis` and an updated hash for the copy, so it probably isn't worth it at this stage.  
```suggestion
  scdblfinder_results <- run_scdblfinder(
    sce,
    cores = opts$cores,
    # only used if there are multiple samples, e.g. if this is processing a merged object
    sample_var = opts$sample_var, # will be NULL if not provided as input argument
    random_seed = opts$random_seed,
    columns_to_keep = keep_columns
  )
  readr::write_tsv(scdblfinder_results, opts$output_file)
```

jashapiro (approved):
I just ran this locally, with test data and it seemed to go fine, so I expect it to work now with the updated container.

I suggested some very minor whitespace updates, which if you accept them should trigger the rerun for confirmation. My other suggestion should probably be ignored for the reasons I state in the comment itself.

modules/doublet-detection/resources/usr/bin/run_scdblfinder.R
sjspielman: ignoring per previous comment

PR #75, opened 2024-07-17
70 Add notes about tags and releases

jashapiro:
I added some brief notes here about code and data releases. 

I realized while writing that there was no particular reason that we had to use fake semantic tags for dated releases, so I changed to giving those just ISO date format, allowing those to also trigger the workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The following base command will run the main workflow, assuming all AWS permissi
 nextflow run AlexsLemonade/OpenScPCA-nf -profile batch
 ```
 
-For most use cases you will want to use the `--results_bucket` argument to avoid writing to the default output bucket.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ The following base command will run the main workflow, assuming all AWS permissi
 nextflow run AlexsLemonade/OpenScPCA-nf -profile batch
 ```
 
-For most use cases you will want to use the `--results_bucket` argument to avoid writing to the default output bucket.
+For many use cases you may want to use the `--results_bucket` argument to avoid writing to the default output bucket (the default is a staging bucket, but may not be acecssible to all users).
```
jaclyn-taroni, line 21:
```suggestion
For many use cases you may want to use the `--results_bucket` argument to avoid writing to the default output bucket (the default is a staging bucket, but may not be accessible to all users).
```

jaclyn-taroni (approved):
Spotted a typo, but LGTM otherwise

Issue #73, opened 2024-07-17
Add docker image tags to all docker images


sjspielman:
> Is there another tagged version we can use? The answer may be no for now, but we will want to have tagged versions where we can.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-nf/pull/72#discussion_r1681327974_
           
----

Eventually, we'll want a better (non-latest) tag for the `doublet-detection` docker image.

jashapiro:
I updated this because I think we want to make sure all of the images have versioned tags. This should be possible after https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/623 is completed.

The other module that is affected by the OpenScPCA-analysis release is `simulate-sce`.

The `scpcatools` images in `merge-sce` can be set to 0.4.0, the latest release.

PR #76, opened 2024-07-18
Set scpcatools docker images to v0.4.0

jashapiro:
As part of the expanded #73, I am setting the docker image version for `scpcatools` packages as used by the merge workflow.


sjspielman (approved):
LGTM, these indeed appear to be the only times `scpcatools` images are used 👍

PR #77, opened 2024-07-24
set docker images to v0.1.0

### LINKED ISSUE #73: Add docker image tags to all docker images

jashapiro:
Does what it says. No more `latest` tags!

closes #73

### DIFF
--- a/modules/doublet-detection/main.nf
+++ b/modules/doublet-detection/main.nf
@@ -2,7 +2,7 @@
 
 // Workflow to detect doublets in a SingleCellExperiment object using scDblFinder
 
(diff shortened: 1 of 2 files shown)


sjspielman (approved):
🚀

PR #78, opened 2024-07-25
Fix S3 upload error

jashapiro:
The S3 upload error we were seeing in slack was not actually an upload error, but rather an error in the `rm` step. The log files were being uploaded as expected to S3, but I had erroneously added matching quotes around the glob going into `rm` so it was not being expanded and did not match any files.


sjspielman (approved):
👍

Issue #80, opened 2024-07-26
Create a release: v0.1.0


jashapiro:
If issues must be resolved before creating a release, mark them as blockers in ZenHub.

**Pre-release checklist:**

- [x] Add the upcoming release version to add to the title of this issue!
- [ ] All blocking issues have been resolved
- [x] All Docker images use a versioned tag (i.e. `:latest` or `:edge` is not used)
- [x] The default data release in [`nextflow.config`](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/nextflow.config) is up-to-date. (See [OpenScPCA-analysis announcements](https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/categories/announcements?discussions_q=category:Announcements) for the latest data release version, or run `./download-data.py --list-releases` from the `OpenScPCA-analysis` repository.)
- [x] The full workflow has been run successfully
  - [x] Trigger with [Run Workflow on AWS Batch](https://github.com/AlexsLemonade/OpenScPCA-nf/actions/workflows/run-batch.yml) GitHub Action using **Workflow run mode** "full" and **Workflow output mode** "staging"
  - [x] Check output data in the staging buckets:
    - [ ] `s3://openscpca-nf-workflow-results-staging`
    - [ ] `s3://openscpca-test-data-release-staging`
    - [x] `s3://openscpca-test-workflow-results-staging`
- [x] Any mentions of the workflow version in the repository have been updated, including
  -  [x] [`nextflow.config`](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/nextflow.config) manifest
  -  [x] [`CHANGELOG.md`](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/CHANGELOG.md) (see more below)
- [x] Write release notes and add them to [`CHANGELOG.md`](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/CHANGELOG.md), which should include the following:
  - Which modules have been added or removed, if any?
  - Have there been any changes to the workflow configuration or launch instructions?
  - What has changed in the workflow documentation, if anything?
- [x] [Create a release on GitHub](https://github.com/AlexsLemonade/OpenScPCA-nf/releases/new) with a new tag of the version number in vX.Y.Z format
  - [x] Populate the description with the release notes added to the changelog. Note that you may find it helpful to use GitHub's automated release notes generation when updating the changelog.

jashapiro:
[...]

PR #81, opened 2024-07-26
add version numbers for v0.1.0

jashapiro:
That's it! Since the release notes were already largely written in the last PR, I think the only thing left to do was to add a version number. 

If there is anything else you think we need before #80 and the release, let me know!


sjspielman (approved):
🚀 

> If there is anything else you think we need before https://github.com/AlexsLemonade/OpenScPCA-nf/issues/80 and the release, let me know!

I have an urge to check off these boxes `The full workflow has been run successfully` as a treat, but I'll let you handle your own issue there :)

PR #82, opened 2024-07-26
Increase default memory for run_scdblfinder

jashapiro:
We were getting a lot of failures for out of memory errors for `run_scdblfinder`, so let's bump up the default memory to 8 GB.


sjspielman (approved):
👍

PR #86, opened 2024-07-29
mask tokens in logs

jashapiro:
Since nextflow leaks tower tokens (see https://github.com/nextflow-io/nextflow/issues/5183) to the log when making wave requests (and possibly elsewhere?) I am here adding `sed` steps to mask all of those. It only masks the current token value, because I didn't yet figure out what the full pattern for the Seqera tokens is. 

I also moved the workspace ID to 1Password, just to have both in the same place. AWS secrets have been updated to include that secret path as well.


jaclyn-taroni (approved):
👍🏻

jashapiro:
On a positive note, the Nextflow developers are quite responsive, and the root cause should be fixed in the next release: https://github.com/nextflow-io/nextflow/issues/5183#issuecomment-2257707167

PR #93, opened 2024-08-19
Add nextflow -resume setup on workflow runs

### LINKED ISSUE #92: Allow nextflow -resume on workflow runs
While we work to debug the container pulling issues of #85, it might be nice to allow resuming of workflows to prevent having to repeat everything when jobs start to fail

We can make this a launch option used during workflow dispatch and passing that option along to the launch script. 

I’m not sure whether this should be the default for automatically triggered workflows, but my inclination is that it should be. If the input data or process has changed, the workflow should reprocess automatically, and we may be able to save ourselves from some unnecessary repeat processing when rerunning due to transient or throttling failures.

jashapiro:
Closes #92

Here I am adding the option to use the `-resume` flag with the Nextflow runs. It should be set as an option on manual runs, but I am actually setting it by default in the automated deploy runs, which I think should be fine, but if it turns out to be a problem, we can always turn it off there as well.

### DIFF
--- a/.github/workflows/run-batch.yml
+++ b/.github/workflows/run-batch.yml
@@ -30,6 +30,10 @@ on:
         options:
           - staging
           - prod
+      resume:
+        description: Use -resume flag for Nextflow launch
+        type: boolean
+        default: true
 
 permissions:
   id-token: write # This is required for requesting the JWT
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
👍🏻

Issue #85, opened 2024-07-29
Wave containers not being pulled


jashapiro:
When trying to deploy the v0.1.0 and v0.1.1 releases, we ran into errors where containers were not being pulled as expected, leading to repeated failures. Test runs do just fine, but the later runs started to fail, which seems to suggest that there might still be an issue with rate limits related to wave containers. 

While we are populating the `TOWER_ACCESS_TOKEN` environment variable on launch, and we know this is getting accepted because we can monitor the runs on Seqera cloud, I wonder if we might actually need to populate `tower.accessToken` configuration variable as well/instead for wave containers specifically. While this would surprise me, I can't think of another reason at the moment. 

There is also a possibility that there is an error somehow related to wave pulling containers from ECR, though again this seems unlikely to be a root cause. 

My plan is to wait and try to deploy the 0.1.1 release one more time, but to first have a more careful look at the log files to try to determine if there is something more specific that I can find. 

If this can not be resolved, we may need to abandon the use of wave containers, but I am hoping it does not come to that, and I will try to ask Seqera for assistance before we get to that stage.

jashapiro:
I updated `nextflow` to on the AWS workload machine to the latest version, 24.04.3, and triggered a manual run for workflow version v0.1.1 (only the real data, as the simulations had gone fine). That is currently running at https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/4Rrf1pyhMkOBw6 and seems to be working as expected, pulling images as expected.

Hoping that there was a bug in the previous nextflow version, though I can't rule out that the time delay might mean that we have gotten past any throttle limits before I began the run.

I did check the previous logs and confirmed that the token was being sent as expected.

jashapiro:
The previous run is on track to complete, with only merge tasks remaining, despite a series of apparent failures for container pulls (zero execution time) during doublet detection. I am still not sure what is going on exactly, but I will keep this open for monitoring.

If we are running into pull limits, one solution may be to batch samples together to reduce the number of jobs.

Alternatively, if we can get a way to use the Fusion file system without relying on so heavily on the Wave system, that might be ideal.

jashapiro:
I take it back. The job seems to have stalled, and we are still seeing container pull errors. I'm still not sure what is going on.

jashapiro:
Looking at the actual error on AWS Batch, I see:

```
[...]

PR #94, opened 2024-09-10
Minor updates to configs for local & batch

jashapiro:
As described in https://github.com/AlexsLemonade/OpenScPCA-nf/issues/85#issuecomment-2341765247, I discovered a few small setting changes while testing why we were getting errors pulling containers, and I am adding them here:

The first is the removal of `scratch = false` in favor of the default setting. I think that we had originally set this to accommodate the use of a dynamically expanding volume which did not include the scratch directory in `scpca-nf`, but with the `fusion` file system this should not be needed.

The other change is to make the local run use the correct platform on Apple Silicon and remove a deprecated option.

When this PR goes in, I will run another test to see if #85 seems to really be resolved.

### DIFF
--- a/config/profile_batch.config
+++ b/config/profile_batch.config
@@ -5,7 +5,6 @@ docker.enabled = true
 wave.enabled = true
 fusion.enabled = true
 
-
 aws {
   batch{
     maxTransferAttempts = 3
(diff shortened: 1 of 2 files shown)


sjspielman (approved):
👍 , presumably!

PR #100, opened 2024-11-12
Update anndata formatting

### LINKED ISSUE #99: Update simulations to match scpca-nf v0.8.5
The simulated data here has drifted a bit from the scpca-nf output for h5ad files, so updates are needed here.

Originally I thought this only affected the names of the reduced dimension variables, but there are a few more changes that were made, so more substantial updates are needed. 

The main approach will be to adapt and use [reformat_anndata.py](https://github.com/AlexsLemonade/scpca-nf/blob/main/bin/reformat_anndata.py) in the simulate-sce workflow and update other components as needed.

jashapiro:
Closes #99

Here I am adding in a modified version of the anndata reformatting script from scpca-nf to bring the formatting of the simulated data up to date with the published data.

The main code of the script is the same, but wrapped into a function so that I can apply it to all files in a directory.

I also updated the initial export script so it includes the PCA metadata output that goes into the h5ad files. 

I have tested this with local data, and things seemed to go well, but a full test still needs to be done. (Starting it after filing this PR)


jashapiro:
Can confirm that the workflow completed without errors: output from the simulation step (SCE & AnnData) can be found at s3://openscpca-test-data-release-staging/test/

modules/simulate-sce/resources/usr/bin/reformat_anndata.py
```diff
@@ -0,0 +1,79 @@
...
+        "--dir",
+        help="directory containing H5AD files and PCA metadaa",
+        required=True,
+    )
+
+    args = parser.parse_args()
+
+    # find all processed rna h5ad files in the directory, recursively
+    anndata_files = list(pathlib.Path(args.dir).rglob("*_processed_rna.h5ad"))
+    pca_files = [str(f).replace("_rna.h5ad", "_pca.tsv") for f in anndata_files]
```
allyhawkins, line 76: Do we have cases with the simulated data where the PCA can't be computed for any reason (too few cells?) and would that break this if there wasn't a 1:1 of processed H5 and PC matrix files? If that's a possibility, I think it's worth making sure we are matching the files by sample ID rather than just pairing the items together based on file order.

modules/simulate-sce/resources/usr/bin/reformat_anndata.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+import argparse
+import pathlib
+
+import anndata
+import pandas as pd
+
+
+def reformat_anndata(anndata_file, pca_metafile):
```
allyhawkins, line 20: If we do think we could have a missing PCA file, we should modify this function to accommodate that.

allyhawkins (commented):
This looks mostly good, I just had one thought about ensuring this script would work for cases where we don't have a PCA matrix. I don't know how relevant that is with the simulated data though.

modules/simulate-sce/resources/usr/bin/reformat_anndata.py
jashapiro: The function handles this at line 39: It checks whether the PCA file exists and skips the related sections in that case.

modules/simulate-sce/resources/usr/bin/reformat_anndata.py
jashapiro: Yes, here we are just making the expected PCA file path; if the files doesn't exist, we will skip the PCA processing.

allyhawkins (approved):
Thanks for answering my earlier questions, LGTM!

PR #102, opened 2024-11-14
Update metadata to include project-specific fields

### LINKED ISSUE #101: metadata fields dropped during permutation
Arising from https://github.com/AlexsLemonade/OpenScPCA-admin/issues/291

The `permute-metadata.R` script in the `simulate-sce` module specifies a set of metadata fields that need to be handled, but the `single_cell_metadata.tsv` file that is uses as input can include more columns from the submitter's sample info, and these should be retained. 

This fix should really be done first in `OpenScPCA-analysis` and then ported to here. I will link that issue here once I file it.

jashapiro:
Here I am porting over the changes from https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/879 that update the simulated metadata files to include project-specific fields. I also updated the readme to reflect the new script source commit and the CHANGELOG to mention these changes.

closes #101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ You may want to add temporary notes here for tracking as features are added, bef
 - Update simulations to match current (v0.8.5) `scpca-nf` output
   - Change reduced dimension names in AnnData output (to `X_pca` and `X_umap`) and updated formatting to match scpca-nf v0.8.5
   - Use new age columns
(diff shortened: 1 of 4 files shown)


jashapiro:
Running with fix at https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/5tGQr3j4EjODHs

jashapiro:
@sjspielman This should be ready for review now.

PR #109, opened 2024-12-19
Add module for seurat conversion

### LINKED ISSUE #108: Add module for converting SCEs to Seurat
Following https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/945, we will want to port the script for converting libraries to Seurat format and publish those files to the results bucket.

For now, this module will only convert `processed` libraries.

jashapiro:
closes #108

Here I am adding a module to convert SCE objects to Seurat format. This directly uses the R script from corresponding OpenScPCA-analysis module, and the associated docker container. 

At the moment, I am only converting the processed data objects; we could also do filtered if desired, but it might make sense to do that as a separate workflow, or to have separate emissions for the workflow? If we want to do both, I will consider options for that conversion, but I expect most usage to be the processed objects for now.

I am labeling this as a draft in part because I think we will need a release of OpenScPCA-analysis so that we can use a versioned Docker image; right now I am just using `latest`, but that is not what we will want for the final workflow. 

With this PR filed, I will run some tests on the AWS infrastructure, which I will report on in comments.

### DIFF
--- a/modules/seurat-conversion/README.md
+++ b/modules/seurat-conversion/README.md
@@ -1,0 +1,7 @@
+This module converts SingleCellExperiment objects to Seurat format.
+
+Scripts are derived from the the `seurat-conversion` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
+
+Links to specific original files used in this module:
+
+- `convert-to-seurat.R`:  <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/27d1947869f3698c3b20a4481670fb00ea984a20/analyses/seurat-conversion/scripts/convert-to-seurat.R>
--- a/modules/seurat-conversion/README.md
+++ b/modules/seurat-conversion/README.md
@@ -1,7 +1,9 @@
 This module converts SingleCellExperiment objects to Seurat format.
 
+Currently, only processed SCE files are converted.
+
 Scripts are derived from the the `seurat-conversion` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
 
 Links to specific original files used in this module:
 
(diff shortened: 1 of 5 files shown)


jashapiro:
Looks like this _almost_ works for the simulated data, with an error for `SCPCS000144/SCPCL000531_processed.rds` that may be related to a small number of cells. 

https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/3BaYaaRw8DExQ

I expect the changes require changing the upstream code; testing to come.

jashapiro:
Run on test data was successful: https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/3x0o3WeAmxbWcY

jashapiro:
run with real data: in progress, but expected to finish successfully: https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/2Xa7g3W4qOv3gW

jashapiro:
The run with real data appears to have been successful. Results can be found in the staging bucket: `s3://openscpca-nf-workflow-results-staging/2024-11-25/seurat/`

modules/seurat-conversion/README.md
```diff
@@ -0,0 +1,7 @@
+This module converts SingleCellExperiment objects to Seurat format.
+
+Scripts are derived from the the `seurat-conversion` module of the [OpenScPCA-analysis](https://github.com/AlexsLemonade/OpenScPCA-analysis) repository.
+
+Links to specific original files used in this module:
+
+- `convert-to-seurat.R`:  <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/27d1947869f3698c3b20a4481670fb00ea984a20/analyses/seurat-conversion/scripts/convert-to-seurat.R>
```
sjspielman, line 7:
```suggestion
- `convert-to-seurat.R`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/27d1947869f3698c3b20a4481670fb00ea984a20/analyses/seurat-conversion/scripts/convert-to-seurat.R>
```

sjspielman (approved):
LGTM! I also downloaded a couple files from the results bucket, and objects looked at expected to me. 

One comment - do we plan to run this on filtered data imminently? If not, we should probably add to the module's README that only processed objects are currently considered in the workflow.

PR #113, opened 2025-01-14
Add module for assigning consensus cell types

### LINKED ISSUE #111: Port module for assigning consensus cell types from OpenScPCA-analysis
We want to add a module for assigning consensus cell types to all ScPCA samples, using the [consensus cell type reference](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv) that we created in `OpenScPCA-analysis`. We should use the same approach used in [`assign-consensus-celltypes.sh`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/assign-consensus-celltypes.sh). 

To complete this, we need to do the following: 

- [ ] move the [reference files](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-consensus/references) to somewhere on S3 that can be used as input to this module. 
- [ ] Port over the scripts, [`03-save-coldata.R`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/scripts/03-save-coldata.R) and [`04-combine-celltypes.R`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R)
- [ ] write a workflow that has two processes, one for grabbing the colData and one for combining everything and assigning the consensus labels.

allyhawkins:
Closes #111 

This PR adds a new module for assigning consensus cell types. I pretty much just moved everything run in [`assign-consensus-celltypes.sh`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/assign-consensus-celltypes.sh) to a Nextflow module here. 

- The first step is run for each sample and grabs the cell type annotations from the `colData` and saves it to a TSV file. Note that this process does not include a publish step since I don't think we really need these tables for anything. 
- The second step combines all tables for all libraries in a single project and assigns the consensus labels to all cells in that project. This means that instead of having a single TSV file for all ScPCA samples we have one for each project. I think that's okay and keeps the overall file size more manageable. I could also imagine a scenario where we might want to use just the results from a single or subset of projects. 

The content of the scripts are the same as the ones that live in `OpenScPCA-analysis` except for: 
- A small change to let the combine script take in a list of files rather than a directory. 
- Accounting for libraries that don't have cell types (cell line samples) by writing an empty file and then removing the empty file before reading in and combining the data frames. 

I also did end up using the permalinks for the reference files. I think right now we want to be able to keep track of any potential version changes, but I don't have a super strong preference either way so could be convinced to use `main` instead. 

I'm currently testing this so will be sure to comment with results from testing.


modules/cell-type-consensus/main.nf
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env nextflow
+
+// Workflow to assign consensus cell type labels
+
+// module parameters
+params.panglao_ref_file = file('https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv')
+params.consensus_ref_file = file('https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv')
```
jashapiro, line 7: In theory, module params are supposed to be deprecated.

We should probably move these to a separate `module_params.config` file which is imported in `nextflow.config`. There is one other module that does this, so we could move those params as well. 

We will probably also want to rename the parameters a bit to be more specific to their content. perhaps `cell_type_consensus_ref` or something like that?

modules/cell-type-consensus/main.nf
```diff
@@ -0,0 +1,103 @@
...
+  output:
+    path consensus_output_file
+  script:
+    input_files = cell_type_files.join(',')
+    consensus_output_file = "${project_id}_consensus-cell-types.tsv.gz"
+    """
+    combine-celltype-tables.R \
+      --input_tsv_files ${input_files} \
+      --panglao_ref_file ${params.panglao_ref_file} \
+      --consensus_ref_file ${params.consensus_ref_file} \
```
jashapiro, line 63: Using the inputs from above. 
```suggestion
      --panglao_ref_file ${panglao_ref} \
      --consensus_ref_file ${consensus_ref} \
```

modules/cell-type-consensus/main.nf
```diff
@@ -0,0 +1,103 @@
...
+  container params.consensus_cell_type_container
+  tag "${sample_id}"
+  errorStrategy 'terminate'
+  input:
+    tuple val(sample_id),
+          val(project_id),
+          path(library_files)
+  output:
+    tuple val(project_id),
+          path(output_files)
```
jashapiro, line 19: For consistency, I think it might be better to still include the `sample_id` here. I know you are not using it, but the consistent expectations has some value.
```suggestion
    tuple val(sample_id),
          val(project_id),
          path(output_files)
```

modules/cell-type-consensus/main.nf
```diff
@@ -0,0 +1,103 @@
...
+        def library_files = Utils.getLibraryFiles(sample_path, format: "sce", process_level: "processed")
+        return [sample_id, project_id, library_files]
+      }
+
+    // save cell type information for each library
+    save_celltypes(libraries_ch)
+
+    cell_type_files_ch = save_celltypes.out
+      .groupTuple(by: 0) // group by project id
+      .map{ project_id, celltype_files -> tuple(
```
jashapiro, line 93: If you take my suggestion to leave the `sample_id` as part of the output, we need a few changes here. 
```suggestion
      .groupTuple(by: 1) // group by project id
      .map{sample_ids, project_id, celltype_files -> tuple(
```

jashapiro (commented):
Overall this looks good! I had a few suggestions, one of which (passing in the files to the process) may actually solve the last error I saw.

I also should also say, you should also be able to some initial testing locally if it helps your development.
- Use `-profile standard,simulated` to run locally with only simulated data, and you shouldn't even need an aws profile to get the data files, as they are public.
- If you add `--project SCPCP000001` or similar you can run just one project for testing
- I realized that you can also force only the module you need with 
- The one caveat is you will probably need to pull your docker image first manually with the `--platform linux/amd64` flag, as Nextflow won't add that itself. 

This is from memory last time I did it, so come back to me if that doesn't work. If it does work, I will add it to the internal readme.

jashapiro:
> * The second step combines all tables for all libraries in a single project and assigns the consensus labels to all cells in that project. This means that instead of having a single TSV file for all ScPCA samples we have one for each project. I think that's okay and keeps the overall file size more manageable. I could also imagine a scenario where we might want to use just the results from a single or subset of projects.

I actually was wondering whether we want to keep the results here at the sample level. If downstream analyses want to use these results, it will often be easier to do that with sample level results. Is there a reason we would want to consolidate these to project level (other than convenience for analysis)?

allyhawkins:
> I actually was wondering whether we want to keep the results here at the sample level. If downstream analyses want to use these results, it will often be easier to do that with sample level results. Is there a reason we would want to consolidate these to project level (other than convenience for analysis)?

Honestly, I don't think I had a real reason for trying to combine results from each sample. I think originally when this wasn't going to live in Nextflow it was just going to be easier to create one results file and read that in to a notebook to summarize rather than reading in close to 700 individual files. But I don't know that we have a valid reason to not have it just output one file per sample. 

I do think if we made that change there would be some pretty big changes to the scripts themselves. We wouldn't need to break it up into two processes and would just need one script that grabs the cell type annotations from the `colData` and merges with the consensus reference. Do we want to do that? And if we do, we might consider saving the data frame that has the blueprint ontology IDs and mapped ontology names to a separate reference file rather than having to load the full ontology file from the URL and do the mapping for every single sample.

modules/cell-type-consensus/main.nf
allyhawkins: The reason I did this is because of this line in the `porting-modules.md` file: 

> If the workflow emits results at the project level, [project_id, output_files] can be used.

See https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/porting-modules.md#module-output-emit. But this may be a moot point if we change to providing results at the sample level.

allyhawkins:
> * Use `-profile standard,simulated` to run locally with only simulated data, and you shouldn't even need an aws profile to get the data files, as they are public.

I don't think this works without being logged in with an aws profile: 

```
nextflow run main.nf -profile standard,simulated --project SCPCP000001 -with-tower
Nextflow 24.10.3 is available - Please consider updating your version to it

 N E X T F L O W   ~  version 24.04.4

Launching `main.nf` [big_poisson] DSL2 - revision: d79706a86b

ERROR ~ Bad Request (Service: Amazon S3; Status Code: 400; Error Code: 400 Bad Request; Request ID: G0WK620K9W598004; S3 Extended Request ID: 79+Vtl0LB008vxJoGWxAZ6NP9aI8pz5yRZLHbluYkC45x4f4nnjxrzApxiu4SaaIpVecWe5KEv2kqWTgmam3seIJOR11tOAd6aoWIVXU0UA=; Proxy: null)

 -- Check script 'main.nf' at line: 20 or see '.nextflow.log' file for more details

```

modules/cell-type-consensus/main.nf
jashapiro: I was thinking about that for the workflow as a whole, but for the individual process, we preserve flexibility by emitting both.

jashapiro:
> nextflow run main.nf -profile standard,simulated --project SCPCP000001

allyhawkins:
It looks like the way we are grabbing the remote files isn't quite working. The error from the most recent run said it couldn't find a column that does exist in the file ( see https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/4aZMz3sQ7dg1gZ). So I went to the `work` directory and found the path to where the remote file is being staged and when I go to that path and open the file it's an HTML file, not a TSV.

modules/cell-type-consensus/main.nf
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env nextflow
+
+// Workflow to assign consensus cell type labels
+
+// module parameters
+params.panglao_ref_file = file('https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv')
+params.consensus_ref_file = file('https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv')
```
jashapiro, line 7: This should fix the fetching to get the raw files. 
```suggestion
params.panglao_ref_file = file('https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv')
params.consensus_ref_file = file('https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv')
```

modules/cell-type-consensus/resources/usr/bin/save-coldata.R
```diff
@@ -0,0 +1,77 @@
...
+
+# check if cell line since cell lines don't have any cell type assignments
+# account for having more than one sample and a list of sample types
+# all sample types should be the same theoretically
+is_cell_line <- all(metadata(sce)$sample_type == "cell line")
+
+# only create and write table for non-cell line samples
+if (is_cell_line) {
+  # make an empty filtered file
+  file.create(opt$output_file)
```
jashapiro, line 57: I wonder if rather than doing this we should make a table with `NA` values for the cell types?

jashapiro (commented):
Looks good, with a few small modifications. 

Thank you for consolidating the module params. I would like them to start to get a consistent naming scheme though, and I made some suggestions there.

As far as blank files go, I also don't like them! I wonder if it would be better to create tables of `NA` values, which would simplify some of the passing of values. Alternatively, you could output a file with only the header row? If you did keep `NA` values for every cell in cell line samples, you would probably want to modify the recoding as `Unknown` as that seems not quite right for cell line data.

Finally, I do think as far as the workflow goes, it would make sense to keep the processing at the sample level. But I do think that it also makes sense to get this version in and come back to it later with the modifications.

config/module_params.config
```diff
@@ -0,0 +1,14 @@
+// Module specific parameters are stored here defined here as parameters
```
sjspielman, line 1: just chiming in, this phrasing seems redudant?
```suggestion
// Module specific parameters are defined here
```

allyhawkins:
> As far as blank files go, I also don't like them! I wonder if it would be better to create tables of NA values, which would simplify some of the passing of values. Alternatively, you could output a file with only the header row? If you did keep NA values for every cell in cell line samples, you would probably want to modify the recoding as Unknown as that seems not quite right for cell line data.

I went the route of creating a table with `NA` filled in for the cell type columns instead of just the header. That would cause mutate/join issues in some of the code. In doing this I also added a column for `sample_type` and then if `sample_type` is `cell line`, I keep the consensus labels as NA rather than "Unknown". 

> Finally, I do think as far as the workflow goes, it would make sense to keep the processing at the sample level. But I do think that it also makes sense to get this version in and come back to it later with the modifications.

Definietly agree with this and I can file an issue for us to go ahead and make this change. Should we do this before running the workflow through and actually looking at results? I think if we're going to make the change we mine as well do it before we run it, but wanted to check and see what you thought. 

This is also ready for another look @jashapiro.

jashapiro (approved):
LGTM!

> Definietly agree with this and I can file an issue for us to go ahead and make this change. Should we do this before running the workflow through and actually looking at results? I think if we're going to make the change we mine as well do it before we run it, but wanted to check and see what you thought.

I think this is really up to you. I might run it once with output to the `staging` bucket so that you can see the results sooner. I'm also happy for the reconfiguration to wait for a while as long as there is an issue filed for it. 

But if you are worried about having to redo downstream changes and not worried about the timing, you could do the reconfiguration first.

PR #115, opened 2025-01-17
Update output prefix for seurat conversion

jashapiro:
I realized when writing up release notes that the seurat conversion workflow output directory does not match the associated module name, so I am fixing that here. 

I will follow up by moving the files on S3 directly, to avoid having to rerun the whole workflow.


allyhawkins (approved):
LGTM

PR #117, opened 2025-01-28
Add schema, validation, and better testing

### LINKED ISSUE #112: Add instructions for canceling runs
The internal readme does not describe how to cancel a run in progress, which is sometimes desirable. Doing so requires access to the `workloads` account, and a few unusual steps to login to the running EC2 instance. 

The short version is: 
- From the `workloads` account, go to EC2 and locate the `Nextflow-workload` instance. 
- Click the "Connect" button, then choose the "Session Manager" tab and then click "Connect"
- In the terminal that opens, enter teh following commands 
  - `sudo su ec2-user`
  - `tmux a`

If there is an active run, you should then see the Nextflow run in progress (If there is not, you will get a "no sessions" message). To cancel the run, enter `control-c`.

If you wish to exit without canceling the run, you can simply close the tab, or press `control-b` followed by `d` to detach from the `tmux` session. 

Other instructions to include might be where the active log file is for monitoring. I think it should be at `/opt/nextflow/.nextflow.log`

### LINKED ISSUE #116: Add schema parameter validation to workflow
We would like to implement parameter validation as in this workflow, using the `nf-schema` plugin.
https://nextflow-io.github.io/nf-schema/latest/

Completing this will include creating a `nextflow_schema.json` file that describes the parameters for this module, and implemented the parameter validation that `nf-schema` allows.

If this goes well, we can use the implementation as a template for https://github.com/AlexsLemonade/scpca-nf/issues/758

jashapiro:
Closes #116 
Closes #112

This PR does three main things:
- adds a `nextflow_schema.json` file
- adds validation of parameters to the Nextflow workflow based on that file
- adds a new profile for local testing

In addition, I updated instructions about running the workflow and adding modules, and made some updates to the GitHub Action to check the schema file.

The bulk of the changes are the addition of the `nextflow_schema.json` file. While this is unfortunately a bit annoying to review, I would appreciate any suggestions on wording of `description` or `help_text` fields. 

The other thing that would be good to test is the new `testing` profile. The goal here is that this should run the workflow entirely locally using test data, so it should not need any AWS permissions to run.

Please also let me know if any of the added instructions are unclear, or if they need more detail.


porting-modules.md
```diff
@@ -153,6 +154,24 @@ If a module creates multiple output files (e.g., a table of results and an R obj
 
 Where possible, include the `SCPCS` sample id, `SCPCL` library id, or `SCPCP` project id as appropriate in the file name to facilitate searching and filtering.
 
+### Module parameters
+
+If a module requires additional parameters, these should be defined as entries in the `config/module_parameters.config` file.
+You will also need to add them to the `nextflow_schema.json` file, which can be updating using the following command from the root of the repository:
```
allyhawkins, line 160:
```suggestion
You will also need to add them to the `nextflow_schema.json` file, which can be updated using the following command from the root of the repository:
```

porting-modules.md
```diff
@@ -153,6 +154,24 @@ If a module creates multiple output files (e.g., a table of results and an R obj
...
+
+```bash
+nf-core pipelines schema build
+```
+
+This will launch a web editor to add descriptions, help, and validation rules for the new parameters.
+
+If you do not already have `nf-core` installed, you can use the `environment.yml` file in this repository to create a conda environment with the necessary tools:
```
allyhawkins, line 168:
```suggestion
If you do not already have `nf-core` installed, you can use the `environment.yml` file in this repository to create a conda environment with the necessary tools prior to updating the `nextflow_schema.json` file:
```
Maybe just clarify that this has to be done first?

readme-internal.md
```diff
@@ -53,9 +61,14 @@ You must have access to the CCDL workspace to view the runs.
 #### Using the Nextflow-workload instance
 
 All runs are launched on the `Nextflow-workload` instance in the OpenScPCA AWS workloads account.
-You can log in to this instance to monitor the workflow runs using the AWS console.
+You can log in to this instance to monitor _and cancel_ the workflow runs using the AWS console.
+You will first need to have access to the "workload" account, then you can log into the instance using the "Connect" button on the EC2 console.
```
allyhawkins, line 65:
```suggestion
You will first need to have access to the "workloads" account, then use the AWS console to navigate to the EC2 console and log into the instance via the "Connect" button. 
```
Minor, but I think its "workloads" plural. Also updating the order here just for clarity.

allyhawkins (commented):
> WARN: Invalid config manifest attribute `contributors`

Do we expect to see this warning? 

I think this generally looks fine, I just had a few small suggestions. I also was able to successfully start a test run using the instructions.

jashapiro:
> > WARN: Invalid config manifest attribute `contributors`
> 
> Do we expect to see this warning?

What version of Nextflow do you have? I don't know when this changed, but it may have been more recent than your nextflow version.

allyhawkins (approved):
The warning I was seeing was definitely just a Nextflow version issue. LGTM!

PR #119, opened 2025-01-30
Add user bin directories to path in script

jashapiro:
While we had previously had `nextflow` in  `/usr/local/bin` on the deploy box, this makes updating versions quite challenging. So we would like to move it to a user directory, but we need to make sure that is in the path when we run. So this adds user paths to `PATH` within the launch script, to ensure nextflow can be found and executed.


allyhawkins (approved):
LGTM

PR #122, opened 2025-02-07
Prep for v0.1.3

allyhawkins:
Towards #121 

Here I'm updating the version tag, adding a changelog entry, and correcting the containers in the schema file. Is there anything I'm missing in the changelog that we should mention?


jashapiro (approved):
LGTM!

PR #126, opened 2025-02-19
Update link to consensus reference

### LINKED ISSUE #125: Update link for consensus celltype reference
We noticed we were missing neuron as a cell type in the consensus reference and made that update in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1037. We then need to re-run all the samples through the `cell-type-consensus` module here using that updated reference. Before we can do that we need to update the URL for the parameter that grabs that file: 

https://github.com/AlexsLemonade/OpenScPCA-nf/blob/3725aca5b701a8a517a97ed9d3b0cdc04def3f5c/config/module_params.config#L13

We were using a tagged version of the file before, so I was going to wait until we have an updated release tag to use (v0.2.2). So we can do this once we close out https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1038. 

Alternatively, we could just update the link to use `main` or the hash for the commit instead of waiting for a release...

allyhawkins:
Closes #125 

Here I am updating the link to the consensus cell type reference since we made a change to it in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1037. I decided to just use the permalink with a hash rather than a tagged version of the file, since we are blocked by https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1040 before we can create a new release of `OpenScPCA-analysis`. 

I want to be able to update this and then re-run for staging and production so that we ensure the results in the workflow bucket are correct. Are we okay with this approach?

### DIFF
--- a/modules/cell-type-consensus/README.md
+++ b/modules/cell-type-consensus/README.md
@@ -10,4 +10,4 @@ This module also uses the following reference files found in the `OpenScPCA-anal
 
 - `blueprint-mapped-ontologies.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4ffb305a87a54133f398c4445ff92348fcb9d8a/analyses/cell-type-consensus/references/blueprint-mapped-ontologies.tsv>
 - `panglao-cell-type-ontologies.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv>
-- `consensus-cell-type-reference.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4ffb305a87a54133f398c4445ff92348fcb9d8a/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv>
+- `consensus-cell-type-reference.tsv`: <https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/e4dc422dd8ddcdc89d031c289fbd123748db500c/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv>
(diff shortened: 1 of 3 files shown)


modules/cell-type-consensus/README.md
```diff
@@ -10,4 +10,4 @@ This module also uses the following reference files found in the `OpenScPCA-anal
 
 - `blueprint-mapped-ontologies.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4ffb305a87a54133f398c4445ff92348fcb9d8a/analyses/cell-type-consensus/references/blueprint-mapped-ontologies.tsv>
 - `panglao-cell-type-ontologies.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b870a082bc9acd3536c5f8d2d52550d8fe8a4239/analyses/cell-type-consensus/references/panglao-cell-type-ontologies.tsv>
-- `consensus-cell-type-reference.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4ffb305a87a54133f398c4445ff92348fcb9d8a/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv>
+- `consensus-cell-type-reference.tsv`: <https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/e4dc422dd8ddcdc89d031c289fbd123748db500c/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv>
```
jashapiro, line 13: Since this is a readme, let's keep the link to the main page (and maybe update the rest of these to matching tag urls?)
```suggestion
- `consensus-cell-type-reference.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4dc422dd8ddcdc89d031c289fbd123748db500c/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv>
```

jashapiro (approved):
I am fine with this approach, but can you file an issue to update these back to tags when there is a new release of openscpca?

PR #130, opened 2025-02-21
Save marker gene expression for cell-type-consensus module

### LINKED ISSUE #125: Update link for consensus celltype reference

### LINKED ISSUE #128: Port saving gene expression of marker genes to consensus cell type module
In https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1050, we added a step to save the expression of all marker genes being used for creating validation plots to the workflow for assigning consensus cell types. It might be a good idea to port the changes to the `cell-type-consensus` module here so that we can run it on all ScPCA samples easily.

allyhawkins:
Closes #128 
Closes #125 

This PR modifies the `cell-type-consensus` module to also export a table of logcounts for all marker genes used for validating cell types (see https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/validation-markers.tsv). I copied over the script from `OpenScPCA-analysis` to incorporate the new changes for saving gene expression. 

For the workflow itself, I modified it to have two output files, one for consensus and one for marker genes. And then I added a new parameter for the marker gene file. 

While I was here I updated the link for the consensus reference to use a tagged version again. I also had to update the consensus docker image since we are now using `scuttle`. I updated all the tags to v0.2.2 while I was at it. 
 
Once all the images for [`OpenScPCA-analysis` v0.2.2](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/13464897162) are finished building then I'll do a test run of the workflow.

### DIFF
--- a/config/containers.config
+++ b/config/containers.config
@@ -19,5 +19,5 @@ params{
   seurat_conversion_container = 'public.ecr.aws/openscpca/seurat-conversion:v0.2.1'
 
   // cell-type-consensus module
-  consensus_cell_type_container = 'public.ecr.aws/openscpca/cell-type-consensus:v0.2.1'
+  consensus_cell_type_container = 'public.ecr.aws/openscpca/cell-type-consensus:latest'
 }
--- a/config/containers.config
+++ b/config/containers.config
@@ -10,14 +10,14 @@ params{
   scpcatools_anndata_container = "ghcr.io/alexslemonade/scpcatools-anndata:v0.4.1"
 
   // simulate-sce module
-  simulate_sce_container = 'public.ecr.aws/openscpca/simulate-sce:v0.2.1'
+  simulate_sce_container = 'public.ecr.aws/openscpca/simulate-sce:v0.2.2'
 
   // doublet-detection module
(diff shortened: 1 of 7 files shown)


allyhawkins:
Note that the [test run succeeded ](https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/321wLrt87W02kD) and I checked the outputs looked correct in the bucket. This should be ready for review.

Issue #129, opened 2025-02-21
Add cell-type-ewings module


allyhawkins:
Now that we have completed the [`cell-type-ewings` module in `OpenScPCA-analysis`](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings), we want to port over the module including any relevant code that was used to actually assign cell types. 

The module that we add should follow the steps described in the [`Cell type annotation overview`](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings#cell-type-annotation-overview) section of the main README for that module. These steps are ultimately the only steps that were used to get the final assignments. 

Below is a general outline of what the workflow would look like: 

- Run AUCell on each sample using 4 gene sets (2 custom and 2 MsigDB)
  - note that we will need to set a parameter for the `aucMaxRank` which should be equivalent to the value used in the original module
- Assign tumor cells and tumor cell states based on AUC values above/below a set of thresholds. These will also need to be parameters that we set based on values used in the original module 
- Assign normal cells as any cell that does not meet the tumor cell criteria and has an assigned cell type using the consensus labels. All other cells are "unknown"

We probably want to break this up into multiple steps. 

- [ ] Create the initial workflow that only includes running AUCell using input gene sets  
- [ ] Modify the workflow to take consensus cell types as input and assign normal cell types

PR #131, opened 2025-02-27
Initiate module for cell-type-ewings

allyhawkins:
Towards #129 

Here I am initiating the `cell-type-ewings` module and adding the first step, which is to run `AUCell` on every library using a set of gene sets from `MSigDB` and custom gene sets. 

- I copied over the script used to run [`AUCell` on gene signatures in the `cell-type-ewings` module from `OpenScPCA-analysis`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R). 
  - That script was written to be able to accommodate merged objects, which we aren't going to be doing here, so I removed anything pertaining to merged. When actually doing the cell type assignment in `OpenScPCA-analysis`, I did use the results from running `AUCell` on the merged object, but because it's rank based for each cell, we should be able to do the same thing on the individual objects, assuming we use the same thresholds. 
  - We previously were determining the `aucMaxRank` using a percentage of detected genes, rather than providing a number. In the `cell-type-ewings` module we ended up using the `aucMaxRank` equivalent to 1% of the total number of detected genes in the merged object. So I calculated that and set that as the default here, which is 425. 
  - This script used to just quit if there was not enough overlap with the gene sets in the individual SCE, but Nextflow doesn't like not having the output files, so I adjusted it to output a data frame with `NA` values. 
- In this script we run AUCell on a set of gene sets using both custom gene sets and gene sets from MSigDB. 
  - For the MSigDB gene sets, we already were using a reference table with all the gene sets to use as the input, so I kept that as is. 
  - For the custom gene sets, we previously were using an argument that specified a directory that contained gene sets as individual files. Instead of using a directory, I updated this argument to take in a comma separated list of files that contain custom gene sets. I then made this optional, just in case we ever wanted to use this script not with custom gene sets. Honestly, it was probably unnecessary, so I can remove that aspect if you don't like it. 
  - I then added params to use the reference files with MSigDB gene sets and custom gene sets present in `OpenScPCA-analysis`, similar to how we use the consensus cell type reference files. 
 - Finally, I added a workflow to run the module and updated config files with the new params. 

I did test this and things were successful!

### DIFF
--- a/modules/cell-type-ewings/README.md
+++ b/modules/cell-type-ewings/README.md
@@ -8,6 +8,6 @@ Links to specific original scripts used in this module:
 
 This module also uses the following reference files found in the `OpenScPCA-analysis` repository:
 
-- `aynaud-ews-targets.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/gene_signatures/aynaud-ews-targets.tsv>
-- `wrenn-nt5e-genes.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/gene_signatures/wrenn-nt5e-genes.tsv>
-- `msigdb-gene-sets.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv>
+- `aynaud-ews-targets.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/aynaud-ews-targets.tsv>
+- `wrenn-nt5e-genes.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/wrenn-nt5e-genes.tsv>
+- `msigdb-gene-sets.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv>
--- a/modules/cell-type-ewings/README.md
+++ b/modules/cell-type-ewings/README.md
@@ -4,10 +4,10 @@ Scripts are derived from the the `cell-type-ewings` module of the [OpenScPCA-ana
 
 Links to specific original scripts used in this module:
 
-- `01-aucell.R`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R>
+- `01-aucell.R`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/v0.2.2/analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R>
 
 This module also uses the following reference files found in the `OpenScPCA-analysis` repository:
 
-- `aynaud-ews-targets.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/aynaud-ews-targets.tsv>
-- `wrenn-nt5e-genes.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/wrenn-nt5e-genes.tsv>
-- `msigdb-gene-sets.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv>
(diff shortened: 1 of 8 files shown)


allyhawkins:
Hmmm is there a way to get it to skip git hooks here? This change https://github.com/AlexsLemonade/OpenScPCA-nf/pull/131/commits/82ae052ecb25f5591891555461dca06c44eeb81f breaks the code...

jashapiro:
> Hmmm is there a way to get it to skip git hooks here? This change [82ae052](https://github.com/AlexsLemonade/OpenScPCA-nf/commit/82ae052ecb25f5591891555461dca06c44eeb81f) breaks the code...

Oh, that is annoying. What we really want to do here is to add a rule for `typos` that leaves that alone. I will file a PR to this branch with that change (and turning pre-commit back on).

nextflow_schema.json
```diff
@@ -96,6 +96,22 @@
...
           "mimetype": "text/tab-separated-values",
           "description": "Table of marker genes used to validate consensus cell types"
+        },
+        "cell_type_ewings_auc_max_rank": {
+          "type": "integer",
+          "default": 425
+        },
+        "cell_type_ewings_msigdb_list": {
+          "type": "string",
+          "default": "https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv"
```
jashapiro, line 106: Can you add all of the validataion info here and for the other files? I made an attempt at the description for this one, but please update it as needed.. A description for the `max_rank` would also be useful.

```suggestion
          "default": "https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv"
                    "pattern": "\\.tsv$",
          "format": "file-path",
          "mimetype": "text/tab-separated-values",
          "description": "Table of MSigDB gene sets"
```

modules/cell-type-ewings/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Links to specific original scripts used in this module:
+
+- `01-aucell.R`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R>
+
+This module also uses the following reference files found in the `OpenScPCA-analysis` repository:
+
+- `aynaud-ews-targets.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/gene_signatures/aynaud-ews-targets.tsv>
+- `wrenn-nt5e-genes.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/gene_signatures/wrenn-nt5e-genes.tsv>
+- `msigdb-gene-sets.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.1/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv>
```
jashapiro, line 13:
```suggestion
- `aynaud-ews-targets.tsv` : <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/aynaud-ews-targets.tsv>
- `wrenn-nt5e-genes.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/gene_signatures/wrenn-nt5e-genes.tsv>
- `msigdb-gene-sets.tsv`: <https://github.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/msigdb-gene-sets.tsv>
```

modules/cell-type-ewings/main.nf
```diff
@@ -0,0 +1,80 @@
...
+    ewing_aucell(
+      libraries_ch,
+      params.cell_type_ewings_auc_max_rank,
+      file(params.cell_type_ewings_msigdb_list),
+      file(params.cell_type_ewings_ews_high_list),
+      file(params.cell_type_ewings_ews_low_list)
+    )
+
+  emit:
+    ewing_aucell.out // [sample_id, project_id, [list of aucell_output_files]]
```
jashapiro, line 79: Will there be more than one output from this workflow? If so, we should add that as a name in the emit statement:
```suggestion
    aucell = ewing_aucell.out // [sample_id, project_id, [list of aucell_output_files]]
```

jashapiro (commented):
The overall structure here looks good. I had some suggestions about formatting and output.

My other question is about inputs and next steps: is this module also going to use the consensus results? I don't think anything needs to happen with that now, but I was curious, thinking about how the next steps will go.

modules/cell-type-ewings/main.nf
allyhawkins: This is what I'm not totally sure about. I think the answer to this is yes, because I think it could be helpful to output both the AUCell results and the final cell type annotations as two separate TSV files that can be used.

allyhawkins:
> My other question is about inputs and next steps: is this module also going to use the consensus results? I don't think anything needs to happen with that now, but I was curious, thinking about how the next steps will go.

Yes the plan is to use the consensus cell types. I think the next step will be a process that takes as input the consensus cell types and the AUCell results and actually does the cell type assignment. That process will also take in a set of params that define the AUC thresholds to use for classifying cells. The output to that will be a TSV file with the cell type annotations. 

I also updated the validation sections of the schema file, fixed some of the links, and then uncommented out the other modules so this should be ready for another look.

jashapiro (approved):
LGTM

PR #132, opened 2025-02-28
add typos rule for aucThr

jashapiro:
Does what it says! I updated `_typos.toml` to add rules for R files, which in this case is set to preserve `aucThr` as a string.

I also turned back on autofixing.


jashapiro:
hmm... that worked locally, but not on CI, strange

allyhawkins (approved):
LGTM, thanks for making this change!

PR #133, opened 2025-03-05
Add script for assigning cell types to cell-type-ewings module

allyhawkins:
Towards #129 

Here I am adding a script that I plan to run as part of a second process in the `cell-type-ewings` module. The script reads in the AUCell results from the first process, the `cell-type-consensus` results, the SCE object, and the markers used to identify the proliferative tumor cells and uses that information to assign cell types. 

- First, any normal cell types are assigned based on if the consensus annotation exists. If so, the normal annotation is used. 
- The results from AUCell for a specific set of gene sets are then used to define tumor EWS-high and tumor EWS-low. (See more details on exactly what gene sets and thresholds [here](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings#cell-type-annotation-overview)). So we need a way to specify what thresholds to use for which gene sets and which cell states. 
  - One option is to set each of the AUC values as an argument, which would mean having 4 arguments for each threshold. We would need a way to distinguish in the argument name both the cell state and the gene set that the AUC value is used for, e.g., `ews_high_aynaud_targets_auc`, which I didn't love. 
  - Instead of that, I came up with the idea of maintaining a table that has the `cell_state`, `gene_set`, and `auc_threshold` as columns. With this approach we have one argument for the AUC table and then based on the contents of that table, I assign the cell states. I personally like this and think it's easier to maintain, but I wanted to see what you thought about this? Also where would this live? Should I move it to `cell-type-ewings/resources`? 
- Lastly, any cells that have mean expression of the proliferative marker genes > 0 and are tumor EWS-high are relabeled as `tumor EWS-high proliferative`. 

I'm curious what you think about the approach I took here and want to nail down this script first before adding it to Nextflow and figuring out how best to use the output from the `cell-type-consensus` module. That seems like a separate PR.


jashapiro:
> * Instead of that, I came up with the idea of maintaining a table that has the `cell_state`, `gene_set`, and `auc_threshold` as columns. With this approach we have one argument for the AUC table and then based on the contents of that table, I assign the cell states. I personally like this and think it's easier to maintain, but I wanted to see what you thought about this? Also where would this live? Should I move it to `cell-type-ewings/resources`?

Yes, I definitely prefer this option. I think `cell-type-ewings/resources` seems like a perfectly good place for this.

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+    type = "character",
+    default = NULL,
+    help = "Path to TSV file with marker genes to use for labeling EWS-high proliferative cells.
+      Must contain a `cell_state` and `ensembl_gene_id` column.
+      Marker genes that have `cell_state == proliferative` will be used."
+  ),
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "Path to file where results will be saved"
```
jashapiro, line 44:
```suggestion
    help = "Path to file where results will be saved as a tsv"
```

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+    help = "Path to TSV file containing consensus cell type labels for a single library."
+  ),
+  make_option(
+    opt_str = c("--aucell_results_file"),
+    type = "character",
+    default = NULL,
+    help = "Path to TSV file containing the AUC values for a set of gene sets output by aucell.R"
+  ),
+  make_option(
+    opt_str = c("--auc_table"),
```
jashapiro, line 28: for consistency, perhaps:
```suggestion
    opt_str = c("--auc_thresholds_file"),
```

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+# get proliferative marker genes
+proliferative_markers <- cell_state_markers_df |>
+  dplyr::filter(cell_state == "proliferative") |>
+  dplyr::pull(ensembl_gene_id)
+
+# calculate the mean expression of all genes for all cells
+mean_exp <- logcounts(sce[proliferative_markers, ]) |>
+  as.matrix() |>
+  t() |>
+  rowMeans()
```
jashapiro, line 128: This is the only place where you use the SCE in this script. Would it be possible to have this be calculated during as part of the AUCell script or process? It might be nice not to have to pass the SCE files into this process, if we can avoid it.

jashapiro: Separately, any reason this doesn't work? 
```suggestion
mean_exp <- logcounts(sce)[proliferative_markers, ] |>
  colMeans()
```

This is a "works for me" situation, mostly, but I also want to note that I also pulled the row selection to after getting the log counts, to save subsetting the whole object.

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+      dplyr::select(gene_set, auc_threshold) |>
+      tibble::deframe()
+
+    # get cells that meet all criteria, e.g., have auc > threshold for all gene sets
+    cells <- geneset_auc_list |>
+      purrr::imap(\(auc_threshold, name){
+        aucell_results_df |>
+          dplyr::filter(gene_set == name & auc > auc_threshold) |>
+          dplyr::pull(barcodes)
+      }) |>
```
jashapiro, line 106: Pretty minor, but you can avoid the deframe and just use `purrr::pmap()`. I added a little rename to the `select` to make the later comparison unambiguous.

```suggestion
    geneset_auc_df <- auc_threshold_df |>
      dplyr::filter(cell_type == state) |>
      dplyr::select(gs_name = gene_set, auc_threshold)

    # get cells that meet all criteria, e.g., have auc > threshold for all gene sets
    cells <- geneset_auc_df |>
      purrr::pmap(\(gs_name, auc_threshold){
        aucell_results_df |>
          dplyr::filter(gene_set == gs_name & auc > auc_threshold) |>
          dplyr::pull(barcodes)
      }) |>
```

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+      purrr::reduce(intersect)
+
+    # figure out which cells these are
+    cell_state_idx <- which(assignment_df$barcodes %in% cells)
+
+    # label the cells with that cell state
+    # modify the data frame in place with <<-
+    assignment_df$custom_annotation[cell_state_idx] <<- state
```
jashapiro, line 114: I don't love this... (`<<-` in particular)

What if your output for this section was a data frame of `barcode` and `state`?
Then you could `purrr::list_rbind()` those together at the end, followed by joining back to the `annotation_df` in a separate step

One thing this will also allow you to check is whether a cell might get annotated more than once, which I think could theoretically happen?

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
```diff
@@ -0,0 +1,151 @@
...
+# get cells with mean > 0
+proliferative_cells <- names(mean_exp)[mean_exp > 0]
+
+# create final assignments
+assignment_df <- assignment_df |>
+  dplyr::mutate(
+    # assign proliferative based on being EWS-high and mean proliferative > 0
+    custom_annotation = dplyr::if_else(
+      barcodes %in% proliferative_cells & custom_annotation == "tumor EWS-high",
+      "tumor EWS-high proliferative",
```
jashapiro, line 139: Just checking: no other states can be proliferative?

Is there a reason not to just make this more like `paste(custom_annotation, "proliferative")`?

jashapiro (commented):
Overall this looks good. 

I had a concern/suggestion about removing `<<-` which always makes me uncomfortable.

My other thought was whether the proliferative calculations can be made as part of the AUC script, or at least that process. The reasoning here is that we might like to avoid having to load the SCE into two processes, as both copying and loading the objects are slow. 
 
A few other minor suggestions as well.

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
allyhawkins: > I don't love this... (`<<-` in particular)

When I was working on this, I thought "Josh is not going to like this", but was struggling to come up with an alternative route. I'll play around with your solution! 

> One thing this will also allow you to check is whether a cell might get annotated more than once, which I think could theoretically happen?

Hmm I haven't seen this in practice probably because of the way I was assigning in the module. Theoretically, these two states should be distinct, but it's possible this could happen and if so, maybe we just label as "tumor".

modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R
jashapiro: Yeah, I think it would have to be a separate file; I still think that is preferable to passing around the whole object.

allyhawkins: How do you feel about making a second script that calculates the mean expression for the all the custom gene sets in https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/references/tumor-cell-state-markers.tsv, not just proliferative. Then we can run that script as a secondary step within the same process that runs AUCell?

jashapiro: That works for me. While there would be an efficiency gain from not loading the whole object in two different scripts, the modularity is probably worth the loss.

allyhawkins:
Okay I made a few udpates based on comments:

- I created a new script that is used to calculate the mean expression of a given set of marker genes. This script reads in the custom marker gene set file and is run within the same process as AUCell so now we output two files, one with AUC results and one with marker gene expression results. 
- I updated the script for assigning cell types to read in both of those outputs and remove reading in the SCE file. 
- I removed the use of `<<-` and instead made a data frame and re-joined it with the larger data frame. I also accounted for the case where barcodes may be assigned to both High/low states and just left it so both labels would be present. I don't actually think that happens in practice, but just in case. 

This should be ready for another look!

jashapiro (commented):
Looking good. I had some small suggestions, and then one more which will be easy but hard to make as a line comment:

I don't love `custom_annotation` as a name, since it is so generic. What about `ewings_annotation` or something like that?

allyhawkins:
> I don't love custom_annotation as a name, since it is so generic. What about ewings_annotation or something like that?

I was struggling with what to call it. I changed to use `ewing_annotation` and made the other small updates you suggested. I tested things and seems to work as expected.

jashapiro (approved):
LGTM

PR #134, opened 2025-03-07
Ewing cell typing part 3

### LINKED ISSUE #129: Add cell-type-ewings module

allyhawkins:
Closes #129 

This PR adds running the cell type assignment script to the `cell-type-ewings` main workflow and should be the last PR to complete that module. 

- To do this, I had to use the output from the `cell-type-consensus` module as input. I chose to filter to the correct samples in `main.nf` in the root of this repo so that the Ewing module only takes in the correct samples rather than everything. Are we okay with this choice or would we prefer to filter within the Ewing module workflow? 
- I had to make a small change to the script for assigning cell types in testing to account for when no cells are assigned to any of the tumor cell states. 
- I had been struggling with how to match the input files up for the process when accounting for multiple libraries per sample, but ended up dealing with that in bash rather than in groovy. I think I like my approach and it's easy to follow, but let me know if you have an alternative approach you prefer.


jashapiro (commented):
This looks good, and I overall like your approach. I suggested a change to get the library ids with groovy instead though, as it adds a bit more flexibility if file names change for some reason (like your change to not compressing).

I also don't think you need to filter both input channels, but I do think maybe it makes sense to filter in the main workflow? I can't decide, to be honest.

main.nf
```diff
@@ -65,5 +65,6 @@ workflow {
   cell_type_consensus(sample_ch)
 
   // Run the cell type ewings workflow
-  cell_type_ewings(sample_ch)
+  // only runs on SCPCP000015
+  cell_type_ewings(sample_ch, cell_type_consensus.out)
```
jashapiro, line 69: After thinking on it a bit, I think it maybe makes sense to the filtering in the root `main.nf` and remove the filtering from the subworkflow? 
```suggestion
  cell_type_ewings(sample_ch.filter{ it[1] == "SCPCP000015" }, cell_type_consensus.out)
```

modules/cell-type-ewings/main.nf
```diff
@@ -4,7 +4,7 @@
 
 process ewing_aucell {
   container params.cell_type_ewing_container
-  tag "${project_id}"
+  tag "${sample_id}"
```
jashapiro, line 7: Very minor, but can an you also update this in [cell-type-consensus](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/df674ace5e453ec8d090a43b5c420977b361caeb/modules/cell-type-consensus/main.nf#L7-L8)?

modules/cell-type-ewings/main.nf
```diff
@@ -39,31 +40,78 @@ process ewing_aucell {
...
+      consensus_celltype_file=\$(ls ${consensus_celltype_files} | grep "\${library_id}")
+      mean_exp_file=\$(ls ${mean_exp_files} | grep "\${library_id}")
+      aucell_file=\$(ls ${aucell_files} | grep "\${library_id}")
+
+      assign-celltypes.R \
+        --consensus_celltype_file \${consensus_celltype_file} \
+        --aucell_results_file \${aucell_file} \
+        --auc_thresholds_file ${auc_thresholds_file} \
+        --mean_gene_expression_file \${mean_exp_file} \
+        --output_file \${library_id}_ewing-celltype-assignments.tsv)
```
jashapiro, line 103: Fix error that snuck in
```suggestion
        --output_file \${library_id}_ewing-celltype-assignments.tsv
```

modules/cell-type-ewings/main.nf
```diff
@@ -39,31 +40,78 @@ process ewing_aucell {
...
+          path(mean_exp_files),
+          path(consensus_celltype_files)
+    path auc_thresholds_file
+  output:
+    tuple val(sample_id),
+          val(project_id),
+          path(celltype_assignment_output_files)
+  script:
+    library_ids = aucell_files.collect{(it.name =~ /SCPCL\d{6}/)[0]}
+    celltype_assignment_output_files = library_ids.collect{"${it}_ewing-aucell-results.tsv"}
```
jashapiro, line 90: My bad, again.
```suggestion
    celltype_assignment_output_files = library_ids.collect{"${it}_ewing-celltype-assignments.tsv"}
```

allyhawkins:
I added in the suggestions that you made and tested it and things appear to work as planned. This should be ready for another look!

modules/cell-type-ewings/main.nf
```diff
@@ -39,31 +40,76 @@ process ewing_aucell {
...
+        --output_file \${library_id}_ewing-celltype-assignments.tsv
+    done
+    """
+
+  stub:
+    clibrary_ids = aucell_files.collect{(it.name =~ /SCPCL\d{6}/)[0]}
+    celltype_assignment_output_files = library_ids.collect{"${it}_ewing-celltype-assignments.tsv"}
+    """
+    for file in ${aucell_files}; do
+      touch \$(basename \${file%_ewing-aucell-results.tsv}_ewing-celltype-assignments.tsv)
```
jashapiro, line 112: Just making this parallel for more consistent testing
```suggestion
    library_ids = aucell_files.collect{(it.name =~ /SCPCL\d{6}/)[0]}
    celltype_assignment_output_files = library_ids.collect{"${it}_ewing-celltype-assignments.tsv"}
    """
    for library_id in ${library_ids.join(" ")}; do
      touch \${library_id}_ewing-celltype-assignments.tsv
```

modules/cell-type-ewings/main.nf
```diff
@@ -93,6 +138,16 @@ workflow cell_type_ewings {
...
+      .join(consensus_ch, by: [0, 1]) // sample id, project id, aucell, mean exp, consensus, consensus gene exp
+      .map { it.dropRight(1) } // we don't need the consensus gene exp file
+
+    // assign cell types
+    ewing_assign_celltypes(assign_ch, file(params.cell_type_ewings_auc_thresholds_file))
+
   emit:
-    aucell = ewing_aucell.out // [sample_id, project_id, [list of aucell_output_files], [list of marker gene expression output files]]
+    aucell = ewing_aucell.out // [sample_id, project_id, aucell_output_file, mean gene expression output file]
+    celltypes = ewing_assign_celltypes. out // [sample_id, project_id, cell type assignment output file]
```
jashapiro, line 152:
```suggestion
    celltypes = ewing_assign_celltypes.out // [sample_id, project_id, [cell type assignment files]]
```

modules/cell-type-ewings/main.nf
```diff
@@ -93,6 +138,16 @@ workflow cell_type_ewings {
...
+      // join by sample ID and project ID
+      .join(consensus_ch, by: [0, 1]) // sample id, project id, aucell, mean exp, consensus, consensus gene exp
+      .map { it.dropRight(1) } // we don't need the consensus gene exp file
+
+    // assign cell types
+    ewing_assign_celltypes(assign_ch, file(params.cell_type_ewings_auc_thresholds_file))
+
   emit:
-    aucell = ewing_aucell.out // [sample_id, project_id, [list of aucell_output_files], [list of marker gene expression output files]]
+    aucell = ewing_aucell.out // [sample_id, project_id, aucell_output_file, mean gene expression output file]
```
jashapiro, line 151: These are still lists, right?
```suggestion
    aucell = ewing_aucell.out // [sample_id, project_id, [aucell output files], [mean gene expression files]]
```

jashapiro (approved):
LGTM. Some minor suggestions for consistency, but nothing major.

PR #136, opened 2025-03-13
Update data release and add changelog for v0.1.4

allyhawkins:
Towards #135 and related to https://github.com/AlexsLemonade/OpenScPCA-admin/issues/334

Here I'm updating the date for the ScPCA data release to use and then prepping for the tagged release v0.1.4. This included updating the tag in the manifest and then adding a changelog entry. I don't plan on doing this release until next week when the data release is complete. 

Just a note that according to the template in https://github.com/AlexsLemonade/OpenScPCA-admin/issues/334, we will need to create a tag in this repo that is dated to match the data release date. Doing this will trigger the GHA of the full workflow. And then creating a versioned release will also do that, so I think that means running through the full workflow twice. Let me know if I should decouple these changes to account for that, or maybe we don't do a versioned release if we want to avoid running the workflow through twice. We could just change the header for the changelog entry to be the date?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ You may want to add temporary notes here for tracking as features are added, bef
 
 ## v0.1.4
 
-- Default release datae for ScPCA data is set to `2025-03-20`
+- Default release date for ScPCA data is set to `2025-03-20`
 - One new module:
   - `cell-type-ewings`: Assigns cell types to Ewing sarcoma samples in `SCPCP000015`
 - One module as been updated:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ Add new release notes in reverse numerical order (newest first) below this comme
 You may want to add temporary notes here for tracking as features are added, before a new release is ready.
 -->
 
+## v0.1.4
+
+- Default release datae for ScPCA data is set to `2025-03-20`
+- One new module:
+  - `cell-type-ewings`: Assigns cell types to Ewing sarcoma samples in `SCPCP000015`
+- One module as been updated:
+  - `cell-type-consensus`:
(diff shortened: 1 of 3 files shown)


jashapiro:
I think you can just skip the date tag this time around. But we probably want to merge/run this before next week; part of the data release process (and the tagging, in particular) is to ensure that the results bucket has appropriate matching data before the release date passes (in real time).

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ Add new release notes in reverse numerical order (newest first) below this comme
 You may want to add temporary notes here for tracking as features are added, before a new release is ready.
 -->
 
+## v0.1.4
+
+- Default release datae for ScPCA data is set to `2025-03-20`
```
jashapiro, line 13:
```suggestion
- Default release date for ScPCA data is set to `2025-03-20`
```

jashapiro (approved):
LGTM. If you need me to pull the release trigger while you are gone, let me know.

allyhawkins:
> I think you can just skip the date tag this time around. But we probably want to merge/run this before next week; part of the data release process (and the tagging, in particular) is to ensure that the results bucket has appropriate matching data before the release date passes (in real time).

Makes sense to me! And sounds good, I can merge this and run the workflow today as soon as the data is finished copying over.

CHANGELOG.md
jashapiro: Though really, I don't know that we need this in the release notes at all; we don't usually change this?

nextflow.config
```diff
@@ -40,7 +40,7 @@ plugins {
 
 // global default parameters for workflows: output buckets are set to staging by default
 params {
-  release_prefix = "2024-11-25"
+  release_prefix = "2025-03-20"
```
jashapiro, line 43: I just realized this requires an update in the .json file too?

allyhawkins: Done in https://github.com/AlexsLemonade/OpenScPCA-nf/pull/136/commits/92aacba08dafa779f08bd7cf25b2bb4b1ffce884

CHANGELOG.md
allyhawkins: We did in the previous changelog so it seemed like it was worth including?

PR #137, opened 2025-03-13
Update sample type handling

allyhawkins:
Now that we have updated the ScPCA data release, the sample type can be both PDX and cell line. This accounts for that in the consensus cell type module, which I believe is the only module that uses that information.


jashapiro (approved):
LGTM. This will be part of the release as well?

allyhawkins:
> LGTM. This will be part of the release as well?

Yes, the test data failed because of this so I will restart the staging run.

PR #139, opened 2025-05-20
Rename auc_threshold column in AUCell output to avoid duplication

allyhawkins:
It looks like we had an error when assigning tumor cell states in the `cell-type-ewings` module, where instead of using the `auc_threhsold` defined in [`auc-thresholds.tsv`](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/modules/cell-type-ewings/resources/auc-thresholds.tsv) to classify cells, we were accidentally using the pre-calculated AUC threshold determined by `AUCell`. This was because the output of `AUCell` included a data frame with an `auc_threshold` column that provides the automatically determined AUC value that `AUCell` identified. We then read this into a script that assigns cell states, and use a pre-set `auc_threshold` to determine if the cell is classified as tumor cell. But because the column and the variable were both named with `auc_threshold` (see https://github.com/AlexsLemonade/OpenScPCA-nf/blob/a97e88c411cabb1b41e1fce6a7735311fb435051/modules/cell-type-ewings/resources/usr/bin/assign-celltypes.R#L79), it was using the value calculated by `AUCell`. 

This is why the one library, SCPCL000822, looked correct since that library has a nice bimodal distribution and tumor cells are easily classified using the pre-calculated AUC in AUCell. But this doesn't hold true for the other libraries and why we were getting very few tumor cells. 

To fix this, I changed the output from running `AUCell` to have the column be `aucell_threshold` rather than `auc_threshold`. This means no modification to the assignment code to deal with the double variable naming and hopefully makes it more clear that the column contains the threshold determined by `AUCell`.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -53,13 +53,13 @@ workflow {
     .filter{ run_all || it[1] in project_ids }
 
   // Run the merge workflow
-  merge_sce(sample_ch)
+  //merge_sce(sample_ch)
 
   // Run the doublet detection workflow
-  detect_doublets(sample_ch)
+  //detect_doublets(sample_ch)
 
   // Run the seurat conversion workflow
-  seurat_conversion(sample_ch)
+  //seurat_conversion(sample_ch)
 
   // Run the consensus cell type workflow
   cell_type_consensus(sample_ch)
--- a/main.nf
+++ b/main.nf
@@ -53,13 +53,13 @@ workflow {
     .filter{ run_all || it[1] in project_ids }
 
   // Run the merge workflow
-  //merge_sce(sample_ch)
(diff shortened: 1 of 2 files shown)


allyhawkins:
> The explanation and code changes make sense to me! Approving under the assumption that you've locally run this code and it's produced cell types as expected - please let me know if you want me to directly test the code, too.

I'm doing a test run of this and then will grab a few samples and check to make sure that things look as expected. Test run is being run here: https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/4Xt1fi7hSZVDY1/v2/tasks

PR #150, opened 2025-07-16
Add module to create gene order files

### LINKED ISSUE #144: Plan for module to create gene order file
Many OpenScPCA modules use `inferCNV`, which requires a gene order file for input. 
Since this is such a common analysis and commonly-needed file, we plan to implement a module that produces two versions of this file: one for whole chromosomes and one which distinguishes p/q arms (useful in situations where different CNVs are expected between arms). 

For reference we have these existing scripts which make versions of these files; these scripts will not necessarily be directly used here but they can serve as guides:

- Whole chromosomes: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/449bbab168f85bdc5397689c8672f9a0d1cd2e71/analyses/infercnv-consensus-cell-type/scripts/00-make-gene-order-file.R
- Also does arms: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/449bbab168f85bdc5397689c8672f9a0d1cd2e71/analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R

The goal of this issue is to plan out this module. Further issues will be opened to implement it.

sjspielman:
Closes #144 

In planning the inferCNV gene order file module, I ended up just making the module, as it is not a large module! It is running locally and producing files as expected. Filing this as a draft to see how CI feels about it all first.

Briefly:

- I defined 4 module-specific parameters:
  - 2x Input files: GTF and cytoband URIs
  - 2x Output file names for the two gene order files that get created (are there suggestions for a different strategy here besides parameters?)
- The module workflow doesn't `take` anything, and it has one process to just call the script that creates the files. It `emit`s the output files  
- Fixed a couple tiny typos I noticed in other files while I was meandering the repo
- I added a README


config/module_params.config
```diff
@@ -21,4 +21,10 @@ params{
   cell_type_ewings_marker_gene_file = 'https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/tumor-cell-state-markers.tsv'
   cell_type_ewings_auc_thresholds_file = "${projectDir}/modules/cell-type-ewings/resources/auc-thresholds.tsv"
 
+  // infercnv-gene-order-file parameters
+  infercnv_gene_order_file_gtf = 's3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+  infercnv_gene_order_file_cytoband = 'ftp://hgdownload.cse.ucsc.edu/goldenPath/hg38/database/cytoBand.txt.gz'
+  infercnv_gene_order_file_output_no_arms = "inferCNV-gene-order.txt"
+  infercnv_gene_order_file_output_with_arms = "inferCNV-gene-order-chrarms.txt"
```
jashapiro, line 28: I would probably just make these constants in the module, or define them as variables within the module workflow itself. I don't love that they don't have any version info in them; I would want the file name to include that this is for Ensembl 104. So I might have a single `infercnv_gene_order_version` param and use that to define output file names in the workflow. Or better yet pull it from the gtf file name.

I'd expect file names like `inferCNV-gene-order_chr_GRCh38_104.txt` and `inferCNV-gene-order_chr_GRCh38_104.txt` and `inferCNV-gene-order_arms_GRCh38_104.txt` that include the full version info for both the assembly and the annotation. This requires you to enforce expectations about the file name, but I think that is okay.

modules/infercnv-gene-order-file/main.nf
```diff
@@ -0,0 +1,48 @@
...
+workflow infercnv_gene_order_file {
+  main:
+     // Create input channel with URIs to GTF and cytoband files
+    input_ch = Channel.of([
+        file(params.infercnv_gene_order_file_gtf, checkIfExists: false),
+        file(params.infercnv_gene_order_file_cytoband, checkIfExists: false)
+    ])
+
+    // Generate the gene order files
+    create_gene_order_files(input_ch)
```
jashapiro, line 44: I don't know if you actually need a channel here at all. I think you can just do
```suggestion
create_gene_order_files([file(params.infercnv_gene_order_file_gtf), file(params.infercnv_gene_order_file_cytoband)])
```

At which point you can have the input just be two paths, and simplify a bit further to:
```suggestion
create_gene_order_files(file(params.infercnv_gene_order_file_gtf), file(params.infercnv_gene_order_file_cytoband))
```

modules/infercnv-gene-order-file/main.nf
```diff
@@ -0,0 +1,48 @@
...
+
+// Workflow to create gene order files for use with inferCNV
+
+process create_gene_order_files {
+  container params.scpcatools_slim_container
+  label 'mem_8'
+  publishDir "${params.results_bucket}/${params.release_prefix}/infercnv-gene-order-file", mode: 'copy'
+  input:
+    tuple path(gtf_file),
+          path(cytoband_file)
```
jashapiro, line 11: If you don't need a channel for input, there is no reason not to just have these as separate inputs.
```suggestion
  input:
    path gtf_file
    path cytoband_file
```

modules/infercnv-gene-order-file/main.nf
```diff
@@ -0,0 +1,48 @@
...
+  output:
+    tuple path(gene_order_file),
+          path(arms_gene_order_file)
```
jashapiro, line 14: I would make the outputs separate as well, as this can simplify downstream use and allows them to be named with `emit:`
```suggestion
  output:
   path gene_order_file, emit: chromosome
   path arms_gene_order_file, emit: arms
```

modules/infercnv-gene-order-file/main.nf
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env nextflow
+
+// Workflow to create gene order files for use with inferCNV
+
+process create_gene_order_files {
+  container params.scpcatools_slim_container
+  label 'mem_8'
+  publishDir "${params.results_bucket}/${params.release_prefix}/infercnv-gene-order-file", mode: 'copy'
```
jashapiro, line 8: So minor, but a slightly shorter name seems nicer (for the whole module)
```suggestion
  publishDir "${params.results_bucket}/${params.release_prefix}/infercnv-gene-order", mode: 'copy'
```

modules/infercnv-gene-order-file/main.nf
jashapiro: A second thought is that if you have a standard name pattern for the outputs (which could be defined in the script), you can just use that file name (with a glob) to get the files. Something like:

```suggestion
  output:
    path inferCNV-gene-order_chr_*.txt, emit: chromosome
    path inferCNV-gene-order_arms_*.txt, emit: arms
```

to emit `inferCNV-gene-order_chr_GRCh38_104.txt` and `inferCNV-gene-order_arms_GRCh38_104.txt`

modules/infercnv-gene-order-file/resources/usr/bin/prepare-gene-order-files.R
```diff
@@ -0,0 +1,111 @@
...
+  col_names = FALSE
+)
+
+
+# Gene order file with chromosome arms ------------------------------
+
+
+# Load cytoBand file into R and assign column names
+cytoBand <- readr::read_tsv(opts$cytoband_file, col_names = FALSE)
+colnames(cytoBand) <- c("chrom", "chrom_arm_start", "chrom_arm_end", "band", "stain")
```
jashapiro, line 74: Do it in one step!
```suggestion
cytoBand <- readr::read_tsv(
  opts$cytoband_file, 
  col_names =  c("chrom", "chrom_arm_start", "chrom_arm_end", "band", "stain")
)
```

config/module_params.config
```diff
@@ -21,4 +21,10 @@ params{
   cell_type_ewings_marker_gene_file = 'https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/tags/v0.2.2/analyses/cell-type-ewings/references/tumor-cell-state-markers.tsv'
   cell_type_ewings_auc_thresholds_file = "${projectDir}/modules/cell-type-ewings/resources/auc-thresholds.tsv"
 
+  // infercnv-gene-order-file parameters
+  infercnv_gene_order_file_gtf = 's3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
+  infercnv_gene_order_file_cytoband = 'ftp://hgdownload.cse.ucsc.edu/goldenPath/hg38/database/cytoBand.txt.gz'
```
jashapiro, line 26: I wonder if we want these to be workflow-wide parameters? It seems reasonable that the annotation gtf and maybe even the cytoband file could be useful to multiple modules, and we might want to define them once for consistency.

jashapiro (commented):
I didn't look (much) at the script itself, but I left a number of comments about the module & process.

The main thing is that I think we want to include the version info in the output files, but we can otherwise just make the output files have fixed names. This can make the parameters much simpler, and also simplifies downstream expectations.

I'd also probably remove tuples and channels, as described in the line comments. If we ever need to create a set of these files, we can do that with a channel that defines the annotation versions or something like that. But I doubt we will need to do that very soon, and it would probably affect other things at the same time. Which is why I did think we might want to make the gtf input a workflow-wide parameter.

config/module_params.config
sjspielman: > So I might have a single infercnv_gene_order_version param and use that to define output file names in the workflow. Or better yet pull it from the gtf file name.

I like this idea, to pull it from the GTF file name. I think we can assume that basename will be standard enough. Do you have any particular preference for this happening in R vs nextflow?

jashapiro: Not particularly: probably easier to do in R though.

jashapiro (commented):
Overall this looks good, with a few small changes to regex and formatting.

Before I approve, have you tested it on our infrastructure? I think you should be able to do a simulated run (since there is no real data) with the following settings:
 
Screenshot 2025-07-17 at 11 15 08 AM

(I'm mostly asking because I have a fear about some infrastructure permissions that could restrict S3 downloads)

modules/infercnv-gene-order/resources/usr/bin/prepare-gene-order-files.R
```diff
@@ -0,0 +1,114 @@
...
+gtf <- rtracklayer::import(opts$gtf_file, feature.type = "gene")
+
+# prepare gene order data frame
+gene_order_df <- gtf |>
+  as.data.frame() |>
+  # rename to support joining in next section of script
+  dplyr::select(gene_id, chrom = seqnames, gene_start = start, gene_end = end) |>
+  dplyr::mutate(chrom = glue::glue("chr{chrom}")) |>
+  # only keep chr1 - 22 and chrX and chrY
+  dplyr::filter(grepl("^chr([1-9]|1[0-9]|2[0-2]|X|Y)$", chrom))
```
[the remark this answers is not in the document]
sjspielman: I had that in the original script actually, but it got lost in the shuffle! I'll bring it back in.

sjspielman:
> (I'm mostly asking because I have a fear about some infrastructure permissions that could restrict S3 downloads)

This was possibly a healthy fear!
Screenshot 2025-07-17 at 11 56 17 AM

jashapiro:
@stephanie Permissions updated: this should be good to go again (and should run fast with -resume)

jashapiro:
The current error is something odd with pulling the workflow, and I don't know what it is. Investigating.

jashapiro:
I don't know why the pulling appeared, but it seems to be fixed now. And the new part of the workflow did run with the updated permissions! https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/wO97iwp3qPlCN/v2/tasks

nextflow.config
```diff
@@ -45,6 +45,10 @@ params {
   results_bucket = "s3://openscpca-nf-workflow-results-staging"
   sim_bucket = "s3://openscpca-test-data-release-staging"
   project = "all"
+
+  // URIs to reference files
+  gtf_file = 's3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
```
jashapiro, line 50: To get ahead of an upcoming change.
```suggestion
  gtf_file = 's3://scpca-nf-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz'
```

nextflow_schema.json
```diff
@@ -199,5 +199,17 @@
     {
       "$ref": "#/$defs/containers"
     }
-  ]
+  ],
+  "properties": {
+    "gtf_file": {
+      "type": "string",
+      "default": "s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz",
```
jashapiro, line 206:
```suggestion
      "default": "s3://scpca-nf-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz",
```

jashapiro (approved):
LGTM, though there is an upcoming change of the references bucket with the account migration that we might as well get ahead of. When I updated the permissions I was already preparing for this.

Issue #156, opened 2025-07-25
Set seed in AUCell for Ewing module


allyhawkins:
We currently have a parameter to set the seed in the AUCell script run in `cell-type-ewings`, but we don't actually set the seed. We should update that script to set the seed and then it's probably worth it to regenerate results with that seed set. I expect the changes to cell types, if any, to be minor.

allyhawkins:
The seed was set in #157 and results were regenerated in staging and prod to use the new seed.

PR #157, opened 2025-07-25
Set seed in AUCell for Ewing module

allyhawkins:
First step towards #156 

Just setting the seed here and then I will re-run the workflow to regenerate results.


sjspielman (approved):
🐈

PR #171, opened 2025-08-26
Initiate NB module and convert NBAtlas

### LINKED ISSUE #163: Initiate NB annotation module
We have completed the `cell-type-neuroblastoma-04` module, so it's time to port it to this repo. This issue tracks initiating the module along with a couple parameters we'll need:

- posterior probability threshold for calling scANVI: 0.75
- URL for NBAtlas download: `https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/f5969395-5f6e-4c5d-a61a-5894773d0fee/file_downloaded`
- TSV files:
  - validation labels ~; there is already a param for this associated with the consensus module, so we might make this a general parameter~ (edit: I misread that file name. the validation _markers_ file already exists as a parameter but we want the validation _groups_)
  - label map: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/references/nbatlas-label-map.tsv
  - ontology map:  https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/references/nbatlas-ontology-ids.tsv

### LINKED ISSUE #164: Add process and script to format NBAtlas
This issue tracks adding a process and R script to prepare the NBAtlas.
The process should ~curl the atlas file and~ run an R script to prepare SCE and AnnData versions. The script should also export HVGs which will later be used to subset SCE objects when prepping for scANVI annotation.

The script should be based on: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R

edited because @jashapiro reminded me that nextflow is very helpful.

sjspielman:
Closes #163 
Closes #164 

This PR starts the new module which I have aptly named `cell-type-neuroblastoma-04`. It includes:

- Module-specific parameters (some of which have TODOs related to tagging)
  - Note that I overrode the NBAtlas link in the testing profile to run with the subsetted version instead, for time. This will require me to use a testing flag though in other circumstances which I'm not sure how I feel about? The full atlas object just takes a while to download at 2.5 gb (could be worse of course, it's not scimilarity!). I'm really on the fence about bringing in the test version at all, what do you think?
- Added parameters to the schema json; some of the container versions needed to be updated here too, so I went ahead and did that and added some descriptions while I was at al
- Started of the new module
  -  `main.nf` with a process to convert the atlas, and the associated script is there too. The script exports three files for later use: the SCE atlas version, the AnnData atlas version, the list of HVGs for prepping objects for scANVI.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -76,6 +76,8 @@ profiles {
       results_bucket = "test/results"
       sim_bucket = "test/simulated"
       project = "SCPCP000012"
+      // Override this module-specific parameter if testing to use the subsetted NBAtlas instead of the full
+      cell_type_nb_04_nbatlas_url = 'https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/0a569381-3a0c-4eec-863a-e20544b686ed/file_downloaded' // direct link to download `seuratObj_NBAtlas_share_v20240130.rds` from https://doi.org/10.17632/yhcf6787yp.3
     }
     process {
       executor = 'local'
--- a/config/containers.config
+++ b/config/containers.config
@@ -24,4 +24,9 @@ params{
   // cell-type-ewings module
   cell_type_ewing_container = 'public.ecr.aws/openscpca/cell-type-ewings:v0.2.2'
 
+  // cell-type-neuroblastoma-04 module
+  // TODO: Update container tag once available
+  cell_type_nb_04_container = 'public.ecr.aws/openscpca/cell-type-neuroblastoma-04:latest'
+
+
 }
--- a/config/module_params.config
+++ b/config/module_params.config
@@ -22,11 +22,11 @@ params{
   cell_type_ewings_auc_thresholds_file = "${projectDir}/modules/cell-type-ewings/resources/auc-thresholds.tsv"
 
   // cell type neuroblastoma
-  cell_type_nb_scanvi_pp_threshold = 0.75 // posterior probability threshold for assigning cell types using scANVI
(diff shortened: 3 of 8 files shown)


nextflow.config
```diff
@@ -76,6 +76,8 @@ profiles {
       results_bucket = "test/results"
       sim_bucket = "test/simulated"
       project = "SCPCP000012"
+      // Override this module-specific parameter if testing to use the subsetted NBAtlas instead of the full
+      cell_type_nb_04_nbatlas_url = 'https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/0a569381-3a0c-4eec-863a-e20544b686ed/file_downloaded' // direct link to download `seuratObj_NBAtlas_share_v20240130.rds` from https://doi.org/10.17632/yhcf6787yp.3
```
allyhawkins, line 80: I don't think we need to worry about testing here and just use the full model. 2.5 GB really isn't that much.

allyhawkins (commented):
This looks good, but I would remove changing to the subset model for testing. I think that's going to make things too complicated here and it's really not that big.

sjspielman:
> but we probably want to set some dummy defaults in the stub profile so that running the stub check doesn't pull the full reference files. I'm going to add that in with SCimilarity so maybe we should do that here too?

Yeahhhh this makes a lot of sense. I would add a blank file into the module to specify for this, unless you had another strategy you're planning for SCimiliarity?

PR #172, opened 2025-08-26
Module for exporting openscpca annotations

### LINKED ISSUE #161: Add a process for exporting cell type annotations to use as input for scpca-nf
We want to be able to export cell type annotations from cell typing modules to use as input to `scpca-nf` so that we can add these annotations to the processed objects. To do this, we would like to save a single JSON file per library that contains: 

- an array of cell barcodes
- an array of cell type annotations (optionally a second array of ontology IDs) 
- the OpenScPCA-nf version 
- release date for the data used to generate the annotations 
- name of the original module 

To do this, we will need a separate process that takes as input the TSV file with the annotations from the specific module and the names of the columns to save in the annotation array. 

These should be saved to a public bucket and not the existing results bucket so that we can read them in as part of `scpca-nf`. We need to decide if we want to use the existing testing bucket, which is public or create a new public bucket specifically for these results.

allyhawkins:
Closes #161 

Here I'm adding a module to be used for formatting and exporting annotations from OpenScPCA as JSON files to use as input to `scpca-nf`. The new module is named `export_annotations` and takes as input a channel with TSV files from annotation modules. It just runs one process which is an R script to create the JSON file using the TSV file as input. The JSON file includes an array of barcodes, cell type annotations, ontology IDs if present, workflow version, release date and the name of the original analysis module. These files are saved to the annotations bucket within the folder corresponding to that data release, following the same organization as the main workflow results output. 

I made this into a separate module that gets run at the end using inputs from the annotation modules. This is currently just Ewings, but the idea would be to combine all the annotation outputs into a single channel to use as input to this module. Alternatively, I could run this as a separate process at the end of each annotation module and not in the main workflow, but I thought that would be harder to maintain. 

I think the main thing we should figure out here is the approach to passing in the column names for where the annotations are in each TSV. Currently I have it set up so there are two arguments to the script, one for the annotation column and one for the ontology column. This is because right now, we don't have a standard naming convention for what the columns should be named when they are output in the TSV files from an annotation module. For example, the Ewing columns are `ewing_annotation` and `ewing_ontology`. I think there are a few different approaches we could take here: 

- Leave the column names as is and specify what columns to use with arguments as I've done here. This requires us to pass in the column names in the input channel to this workflow. 
- Keep arguments for the column names but make those column names parameters to the workflow rather than hardcoding them in the annotation module. This would require some updates to the Rscript for assigning Ewing cell typing but might be easier long term to have the names of the columns in one place. 
- Require the annotations that we want to port over to be in a specific `openscpca_celltype_annotation` column in the TSV files and remove the arguments alltogether. With this option additional annotation columns can be present, but whatever is getting ported over must be in those columns. 
  - My only hesitation with this option is that these columns don't exist yet for the Ewing data. I would want to add these as additional columns with the existing Ewing output rather than overwrite the columns since there is code in `OpenScPCA-analysis` that uses these annotation columns. 

Another thing I wanted to note is that I think we should set a standard expectation of what the output tuples and TSV files should be from cell type annotation modules so that they can be used as input to format the annotations. Once we decide what that is we should update the docs for porting modules to include that information. 

I did test this and successfully made JSON files that can be easily read into R as a data frame.

### DIFF
--- a/modules/cell-type-ewings/main.nf
+++ b/modules/cell-type-ewings/main.nf
@@ -153,12 +153,14 @@ workflow cell_type_ewings {
         sample_id,
         project_id,
         assignment_files,
-        "ewing_annotation", // annotation column
-        "ewing_ontology", // ontology column
-        "cell-type-ewings" // module name
+        { // annotation metadata
+          module_name: "cell-type-ewings", 
+          annotation_column: "ewing_annotation", 
+          ontology_column: "ewing_ontology" 
+        }
       )}
 
   emit:
     aucell = ewing_aucell.out // [sample_id, project_id, [aucell output files], [mean gene expression files]]
-    celltypes = celltype_output_ch // [sample_id, project_id, [cell type assignment files], annotation column, ontology column, module name]
+    celltypes = celltype_output_ch // [sample_id, project_id, [cell type assignment files], annotation_metadata]
 }
--- a/modules/cell-type-ewings/main.nf
+++ b/modules/cell-type-ewings/main.nf
@@ -147,7 +147,18 @@ workflow cell_type_ewings {
     // assign cell types
     ewing_assign_celltypes(assign_ch, file(params.cell_type_ewings_auc_thresholds_file))
 
+    // add ewing specific metadata to output tuple
+    celltype_output_ch = ewing_assign_celltypes.out
+      .map{ sample_id, project_id, assignment_files -> tuple(
+        sample_id,
+        project_id,
+        assignment_files,
+        "ewing_annotation", // annotation column
+        "ewing_ontology", // ontology column
+        "cell-type-ewings" // module name
+      )}
+
   emit:
     aucell = ewing_aucell.out // [sample_id, project_id, [aucell output files], [mean gene expression files]]
-    celltypes = ewing_assign_celltypes.out // [sample_id, project_id, [cell type assignment files]]
+    celltypes = celltype_output_ch // [sample_id, project_id, [cell type assignment files], annotation column, ontology column, module name]
 }
--- a/modules/cell-type-ewings/main.nf
+++ b/modules/cell-type-ewings/main.nf
@@ -153,11 +153,11 @@ workflow cell_type_ewings {
         sample_id,
         project_id,
         assignment_files,
-        { // annotation metadata
-          module_name: "cell-type-ewings", 
-          annotation_column: "ewing_annotation", 
-          ontology_column: "ewing_ontology" 
-        }
+        [ // annotation metadata
+          module_name: "cell-type-ewings",
+          annotation_column: "ewing_annotation",
+          ontology_column: "ewing_ontology"
+        ]
       )}
 
   emit:
--- a/modules/export-annotations/main.nf
+++ b/modules/export-annotations/main.nf
@@ -11,17 +11,14 @@ process format_annotations {
     tuple val(sample_id),
           val(project_id),
           path(annotations_tsv_files),
-          val(annotation_column),
-          val(ontology_column),
-          val(module_name)
+          val(annotation_metadata)
   output:
     tuple val(sample_id),
           val(project_id),
-          path(json_files)
+          path("*_openscpca-annotations.json")
   script:
     library_ids = annotations_tsv_files.collect{(it.name =~ /SCPCL\d{6}/)[0]}
     json_files = library_ids.collect{"${it}_openscpca-annotations.json"}
-    ontology_included = "${ontology_column}" != "NONE"
     """
     for library_id in ${library_ids.join(" ")};do
       # get the input files for the library id
@@ -29,9 +26,9 @@ process format_annotations {
 
       export-celltype-json.R \
         --annotations_tsv_file \$annotations_file \
-        --annotation_column "${annotation_column}" \
-        ${ontology_included ? "--ontology_column  '${ontology_column}'" : ''} \
-        --module_name ${module_name} \
+        --annotation_column "${annotation_metadata.annotation_column}" \
+        ${annotation_metadata.ontology_column ? "--ontology_column  '${annotation_metadata.ontology_column}'" : ''} \
+        --module_name ${annotation_metadata.module_name} \
         --release_date ${params.release_prefix} \
         --openscpca_nf_version ${workflow.manifest.version} \
         --output_json_file \${library_id}_openscpca-annotations.json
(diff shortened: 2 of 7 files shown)


modules/cell-type-ewings/main.nf
```diff
@@ -147,7 +147,18 @@ workflow cell_type_ewings {
...
+    // add ewing specific metadata to output tuple
+    celltype_output_ch = ewing_assign_celltypes.out
+      .map{ sample_id, project_id, assignment_files -> tuple(
+        sample_id,
+        project_id,
+        assignment_files,
+        "ewing_annotation", // annotation column
+        "ewing_ontology", // ontology column
+        "cell-type-ewings" // module name
+      )}
```
jashapiro, line 159: I don't mind this overall, but I think this may be a good place to add back a `meta`-like structure so we can name info we need here. 
```suggestion
      .map{ sample_id, project_id, assignment_files -> tuple(
        sample_id,
        project_id,
        assignment_files,
        { // annotation metadata
          module_name: "cell-type-ewings", 
          annotation_column: "ewing_annotation", 
          ontology_column: "ewing_ontology" 
        }
      )}
```

modules/cell-type-ewings/main.nf
```diff
@@ -147,7 +147,18 @@ workflow cell_type_ewings {
...
+
   emit:
     aucell = ewing_aucell.out // [sample_id, project_id, [aucell output files], [mean gene expression files]]
-    celltypes = ewing_assign_celltypes.out // [sample_id, project_id, [cell type assignment files]]
+    celltypes = celltype_output_ch // [sample_id, project_id, [cell type assignment files], annotation column, ontology column, module name]
```
jashapiro, line 163: So then the output would be:
```suggestion
    celltypes = celltype_output_ch // [sample_id, project_id, [cell type assignment files], annotation_metadata]
```

modules/export-annotations/main.nf
```diff
@@ -0,0 +1,60 @@
...
+  tag "${sample_id}"
+  label 'mem_8'
+  publishDir "${params.annotations_bucket}/${params.release_prefix}/${project_id}/${sample_id}", mode: 'copy'
+  input:
+    tuple val(sample_id),
+          val(project_id),
+          path(annotations_tsv_files),
+          val(annotation_column),
+          val(ontology_column),
+          val(module_name)
```
jashapiro, line 16: Continuing my thought...
```suggestion
          val(annotation_metadata)
```

modules/export-annotations/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    """
+    for library_id in ${library_ids.join(" ")};do
+      # get the input files for the library id
+      annotations_file=\$(ls ${annotations_tsv_files} | grep "\${library_id}")
+
+      export-celltype-json.R \
+        --annotations_tsv_file \$annotations_file \
+        --annotation_column "${annotation_column}" \
+        ${ontology_included ? "--ontology_column  '${ontology_column}'" : ''} \
+        --module_name ${module_name} \
```
jashapiro, line 34:
```suggestion
        --annotation_column "${annotation_metadata.annotation_column}" \
        ${annotation_metadata.ontology_column ? "--ontology_column  '${annotation_metadata.ontology_column}'" : ''} \
        --module_name ${annotation_metadata.module_name} \
```

modules/export-annotations/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    """
+    for library_id in ${library_ids.join(" ")};do
+      touch \${library_id}_openscpca-annotations.json
+    done
+    """
+}
+
+workflow export_annotations {
+  take:
+    celltype_ch  // [sample_id, project_id, [cell type assignment files], annotation column, ontology column, module name]
```
jashapiro, line 53:
```suggestion
    celltype_ch  // [sample_id, project_id, [cell type assignment files], annotation metadata]
```

modules/export-annotations/main.nf
```diff
@@ -0,0 +1,60 @@
...
+          val(ontology_column),
+          val(module_name)
+  output:
+    tuple val(sample_id),
+          val(project_id),
+          path(json_files)
+  script:
+    library_ids = annotations_tsv_files.collect{(it.name =~ /SCPCL\d{6}/)[0]}
+    json_files = library_ids.collect{"${it}_openscpca-annotations.json"}
+    ontology_included = "${ontology_column}" != "NONE"
```
jashapiro, line 24: I think it makes sense to just use a blank or missing value if there is no ontology column, so I suggested that below.
```suggestion
```

modules/export-annotations/resources/usr/bin/export-celltype-json.R
```diff
@@ -0,0 +1,99 @@
...
+}
+
+# build json contents
+json_contents <- list(
+  barcodes = annotations_df$barcodes,
+  openscpca_celltype_annotation = annotations_df[[opt$annotation_column]],
+  openscpca_celltype_ontology = ontology_ids,
+  module_name = opt$module_name,
+  openscpca_nf_version = opt$openscpca_nf_version,
+  release_date = opt$release_date
```
jashapiro, line 89: Minor, but I'd rather see the module name and such at the top
```suggestion
  module_name = opt$module_name,
  openscpca_nf_version = opt$openscpca_nf_version,
  release_date = opt$release_date,
  barcodes = annotations_df$barcodes,
  openscpca_celltype_annotation = annotations_df[[opt$annotation_column]],
  openscpca_celltype_ontology = ontology_ids
```

modules/export-annotations/main.nf
```diff
@@ -0,0 +1,60 @@
...
+    tuple val(sample_id),
+          val(project_id),
+          path(annotations_tsv_files),
+          val(annotation_column),
+          val(ontology_column),
+          val(module_name)
+  output:
+    tuple val(sample_id),
+          val(project_id),
+          path(json_files)
```
jashapiro, line 20: We could probably save a bit of effort by just letting Nextflow grab all of these. (no need to define `json_files`)
```suggestion
          path("*_openscpca-annotations.json")
```

jashapiro (commented):
Overall this looks good! 

I like keeping the flexibility to name columns flexibly, so having the column names as part of the output seems reasonable to me.
 
My main suggestion is to wrap up the annotation metadata into a tuple to make keeping track of what goes where for the annotation columns a bit easier and more flexible. 

One thought I had at the end was that you could have default values for the columns as well, implemented as something like:
```
--annotation_column "${annotation_metadata.annotation_column ?: "openscpca_celltype_annotation" }" \
```

But now maybe I don't think that adds much, since you will require something in the output anyway...

allyhawkins:
I took the suggestion to wrap the metadata in a tuple, but chose not to add a default for the columns since I want to make sure there actually is a column that should be ported over. Otherwise it doesn't really make sense to have an empty column. 

I also added an annotations bucket for the stub profile otherwise it complains when running the Ewing project. 

I will file an issue to add some information to docs about how to format the output tuples for annotation modules, but otherwise this should be ready for another look.

PR #173, opened 2025-08-27
Add script and process to train SingleR model

### LINKED ISSUE #165: Add process and script to train SingleR model
This issue tracks adding a process and R script to train the SingleR model from the NBAtlas.
The script should be based on:  https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/01_train-singler-model.R

sjspielman:
Closes #165 

This PR adds the process and R script to train the SingleR model for NB annotations. Part of this script entails finding the intersecting gene symbols between the NBAtlas reference and ScPCA objects; previously we did this using an SCE to help, but in this implementation I'm using the GTF file to get those symbols. I also caught/fixed a couple param and other typos while locally testing.

For testing purposes, I'm temporarily emitting this output as well, but this will go away as development proceeds!

### DIFF
--- a/modules/cell-type-neuroblastoma-04/main.nf
+++ b/modules/cell-type-neuroblastoma-04/main.nf
@@ -6,14 +6,12 @@
 process convert_nbatlas {
   container params.cell_type_nb_04_container
   label 'mem_8'
-  // TODO: Remove publishDir after initial PR
-  publishDir "${params.results_bucket}/${params.release_prefix}/cell-type-neuroblastoma-04"
   input:
     path nbatlas_seurat_file
   output:
-    path nbatlas_sce_file
-    path nbatlas_anndata_file
-    path nbatlas_hvg_file
+    tuple path(nbatlas_sce_file),
+          path(nbatlas_anndata_file),
+          path(nbatlas_hvg_file)
   script:
     nbatlas_sce_file = "nbatlas_sce.rds"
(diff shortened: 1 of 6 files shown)


modules/cell-type-neuroblastoma-04/main.nf
```diff
@@ -37,6 +35,31 @@ process convert_nbatlas {
 }
 
 
+process train_singler_model {
+  container params.cell_type_nb_04_container
+  label 'mem_8'
+  publishDir "${params.results_bucket}/${params.release_prefix}/cell-type-neuroblastoma-04"
+  input:
+    path nbatlas_sce_file,
```
allyhawkins, line 43:
```suggestion
    path nbatlas_sce_file
```
This might be your fix.

modules/cell-type-neuroblastoma-04/main.nf
```diff
@@ -37,6 +35,31 @@ process convert_nbatlas {
 }
 
 
+process train_singler_model {
+  container params.cell_type_nb_04_container
+  label 'mem_8'
+  publishDir "${params.results_bucket}/${params.release_prefix}/cell-type-neuroblastoma-04"
```
allyhawkins, line 41: Are you planning to keep the trained model in the final output or is this temporary? I think we could probably remove publishing this.

PR #175, opened 2025-08-27
Add module for running SCimilarity

### LINKED ISSUE #170: Add module for running SCimilarity
We want to annotate all ScPCA samples using [`SCimilarity`](https://genentech.github.io/scimilarity/). This should just be one process that takes as input the processed H5AD files and then outputs a TSV file with annotations following the same setup as the [`run-scimilarity.sh`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-scimilarity/run-scimilarity.sh) script in the `cell-type-scimilarity` module. 

For input, we do need to provide the `SCimilarity` model, which can be found in `s3://scpca-references/celltype/scimilarity_references`. We also need the TSV file mapping ontology IDs to annotations in the [`cell-type-scimilarity` module](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-scimilarity/references/scimilarity-mapped-ontologies.tsv).

allyhawkins:
Closes #170 

Here I'm adding the module to run `SCimilarity` on all of the samples. There's only one step here which is just to run `SCimilarity` and output the annotations as a TSV file. I copied the script that does this from `OpenScPCA-analysis` without any modifications so the main code to review here is the addition to Nextflow. 

- There are two new parameters, one for the model itself and one for the ontology map file that we created in `OpenScPCA-analysis`. Since the model file is quite big, I added an empty folder for stub testing and added that path to the `stub` profile. 
- This module runs on all processed `h5ad` files for RNA only, so I do have a step that should filter out any adt files. 

I am filing this as a draft because I'm having issues getting Nextflow to run the script in the container inside the conda environment. The way that Nextflow launches and runs the image means the conda environment that's installed isn't getting used by default and so it can't find the packages we use in the script. I think I found a solution to when we build the environment to set the default python to the conda environment so that it will work with Nextflow that I'll file as a PR in `OpenScPCA-analysis`. 

Once I'm able to confirm this runs, then I'll request formal review.

### DIFF
--- a/modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
+++ b/modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
@@ -1,4 +1,4 @@
-#!/bin/bash python3
+#!/usr/bin/env python3
 
 # Script to annotate processed ScPCA objects with SCimilarity
 # Follows this tutorial: https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html
--- a/modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
+++ b/modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/bin/bash python3
 
 # Script to annotate processed ScPCA objects with SCimilarity
 # Follows this tutorial: https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html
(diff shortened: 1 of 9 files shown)


allyhawkins:
I ran through the whole workflow with the simulated data successfully, so this is now ready for review.

modules/cell-type-scimilarity/resources/usr/bin/run-scimilarity.py
```diff
@@ -0,0 +1,193 @@
...
+    ################################################
+    ################ Export annotations ############
+    ################################################
+
+    # prepare the predictions with min distance for export
+    predictions_df = pandas.DataFrame(
+        {
+            "barcode": processed_anndata.obs_names.to_list(),
+            "scimilarity_celltype_annotation": predictions.values,
+            "min_dist": nn_stats["min_dist"],
```
sjspielman, line 180: Do we need this column in this workflow?

allyhawkins: Yes! This is a stat we are going to use to measure confidence, as recommended by `SCimilarity` docs, so we want to output it so we can use it for exploratory analysis.

PR #176, opened 2025-08-27
Add process/script to train reference scanvi model

### LINKED ISSUE #167: Add process and script to train the scANVI reference model
This issue tracks adding a process and R script to train a scANVI model on NBAtlas
The script should be based on: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/03a_train-scanvi-model.py

sjspielman:
Closes #167 

This PR adds the last "reference-building" step for the NB workflow, to train the reference scANVI/scArches model. I tested this locally emitting both the singler model file and scanvi _directory_ (scanvi creates a directory with 2 model files in it as its output) ✅ 

While implementing this, I realized I should split up the original `convert_nbatlas` output into 3, since the HVG file and anndata reference file don't actually go into the same process even though they are both scanvi. The anndata reference only goes into the training process, and the hvgs will only go into the forthcoming scanvi label transfer process. So, I tweaked that part as well.

Here are the main differences in the python script compared to its original version:
- Since we don't have to accommodate testing data, we can set the `CELLTYPE_COLUMN` as a constant to the tumor zoom column
- In the context of running this workflow, we can let it automatically choose the accelerator (cpu or gpu), so we don't need to specify an argument (its default accelerator is "auto"). One day if we have gpu, it would therefore be able to run still!
- I only export the model, which is a directory that scvi-tools creates on saving. We don't need to export (or calculate for that matter) the latent representation separately.


allyhawkins (approved):
LGTM!

PR #178, opened 2025-08-27
Add process/script to classify with singler

sjspielman:
Stacked on #176 
Closes #166 

This PR adds the R script and nextflow process to classify samples with SingleR. While working on this, I realized we actually haven't been as comprehensively testing as we maybe want to be (#177), so I'm opening this PR with `SCPCP000004` added as project to run through in the `stub` profile so we can ensure this really is passing CI. I will remove that once this passes to request review.

PR #179, opened 2025-08-28
Classify with scanvi/scarches

### LINKED ISSUE #168: Add process and scripts to classify with scANVI
This issue tracks adding a process and R script to classify NB samples using the ref-trained scANVI model with label transfer.

This process will need two scripts:
- Prepare the query data for input to scANVI, which entails subsetting to HVGs and converting to AnnData. This script should be based on: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/03b_prepare-scanvi-query.R
- Run label transfer The script should be based on: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py

sjspielman:
Stacked on #178 
Closes #168 

This PR adds the scripts (plural!) and process to classify with scANVI/scArches. The two scripts are to i) prepare the input data by converting from sce -> anndata and subsetting to nbatlas' hvgs, and ii) classify cells. The process is about the same as for `classify_singler`, except I pull out the library id to define the intermediate file that gets output from the prep script and input to the classification script. I ran this `stub` process locally including `SCPCP000004` ✅ 

Edit - this has been tested successfully with `simulated` now.

### DIFF
--- a/modules/cell-type-neuroblastoma-04/main.nf
+++ b/modules/cell-type-neuroblastoma-04/main.nf
@@ -137,7 +137,7 @@ process classify_scanvi {
     """
     for file in ${library_files}; do
 
-      anndata_file=\$(basename \${file%.rds}_prepared.h5ad)
+      anndata_file="prepared.h5ad"
       scanvi_tsv=\$(basename \${file%.rds}_scanvi.tsv.gz)
 
       # Prepare the query data for input to scANVI/scArches


sjspielman:
So green! https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/5BUT8vhL5G1FGw/v2/tasks

modules/cell-type-neuroblastoma-04/main.nf
```diff
@@ -118,6 +118,50 @@ process classify_singler {
...
+    path scanvi_ref_model_dir
+  output:
+    tuple val(sample_id),
+          val(project_id),
+          path("*_scanvi.tsv.gz")
+  script:
+    """
+    for file in ${library_files}; do
+
+      anndata_file=\$(basename \${file%.rds}_prepared.h5ad)
```
allyhawkins, line 140:
```suggestion
      anndata_file="prepared.h5ad"
```
You probably don't need to be this fancy and can just define a simpler path since you are just using this to pass the object from one script to the next and not actually outputting this file.

allyhawkins (approved):
This looks good to me, I just had one minor suggestion to simplify things.

PR #180, opened 2025-09-03
Assign neuroblastoma labels

### LINKED ISSUE #169: Add process and script to finalize NB annotations
This issue tracks adding a process and R script to derive final annotations for NB samples.
This script will be based on this notebook, although note we'll be going _per library_ - aka not using the merged object! - in this workflow, so there are some expected code changes in that regard: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/final-annotation.Rmd

sjspielman:
Stacked on #179
Closes #169 

This PR adds the final process to the `cell-type-neuroblastoma-04` module as well as a couple other small changes:

- I appended `nb_04_` to all the processes I had defined, which seems helpful for future-proofing
- I added the new process `nb_04_assign_celltypes` to run the new script `assign-labels.R` which is based on the annotation notebook. Locally, I ran this to ensure this script produces the same final labels on a couple samples as that notebook in the analysis repo did.
- The workflow now emits cell types along with annotation metadata with expected fields
- I mixed the output in with the ewings output to head into the `export_annotations` workflow

I tested this locally with the `testing` profile, but notably I can't test locally _in Docker_ due to memory limitations, so more testing in batch is forthcoming; will request review once that is set!

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -57,17 +57,17 @@ workflow {
     .filter{ run_all || it[1] in project_ids }
 
 
-  // // Run the merge workflow
-  // merge_sce(sample_ch)
+  // Run the merge workflow
+  merge_sce(sample_ch)
 
-  // // Run the doublet detection workflow
-  // detect_doublets(sample_ch)
+  // Run the doublet detection workflow
+  detect_doublets(sample_ch)
 
-  // // Run the seurat conversion workflow
-  // seurat_conversion(sample_ch)
+  // Run the seurat conversion workflow
+  seurat_conversion(sample_ch)
 
-  // // Run the infercnv gene order file workflow
-  // infercnv_gene_order()
+  // Run the infercnv gene order file workflow
+  infercnv_gene_order()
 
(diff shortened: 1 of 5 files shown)


jashapiro:
> * I appended `nb_04_` to all the processes I had defined, which seems helpful for future-proofing

Not opposed to this, but it shouldn't matter. As long as only the workflow is imported into `main.nf` there should not be any conflict with other processes.

sjspielman:
This has run through successfully: https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/1IgqYbSsXzKTce/v2/tasks

I confirmed files are present in the test results bucket and look as expected! I'm attaching one as an example.
[SCPCL000119_neuroblastoma-04_celltype-assignments.tsv.zip](https://github.com/user-attachments/files/22125402/SCPCL000119_neuroblastoma-04_celltype-assignments.tsv.zip)

PR #193, opened 2025-09-29
Add initial support for pull through repositories

jashapiro:
Here I am adding support for using the AWS ECR pull through repository for container pulls. 

The main action here is the addition of a `Utils.pullthroughContainer()` function which takes as arguments the container image address and a new `pullthrough_registry` param. The way it works is that if `params.pullthrough_registry` is defined (not an empty string), the function returns the path to the same container but in the pull through cache. The idea is that if someone else is using this and does not want to use or does not have access to a pull through cache, the workflow will still work just as before, and they can still look at the `containers.config` file to see what the default containers are.

Each source repository has a defined prefix in the pull through cache, which are set in the function itself. If there is not a prefix defined for that repository, we keep the original container as well. At the moment, we define prefixes only for the AWS Public ECR and quay.io, because that is all that is actually set up now in the infrastructure. Other repositories require authentication for a pull through rule, which we have not tackled yet. In practice, this is fine for this repo, as we use module-specific containers, which all live on AWS Public ECR. 

As far as my testing has gone, things seem to be working, but I am filing this as a draft to start because I am waiting to get some of the required infrastructure changes fully fixed over in https://github.com/AlexsLemonade/OpenScPCA-infra/pull/141

As a note: 
I had also tried another approach that would have allowed us to build the  where I planned to let the Seqera Wave service handle the pulling and caching of images using the `wave.freeze` config option, but I ran into trouble where some of our images could not be fully built by the service, as they were too large. Unfortunately, it seems like the addition of the Wave/Fusion components has to be done by their service for freezing images,  and even though the images are already built, I got OOM errors for at least one of the larger images during the build steps there.

### DIFF
--- a/porting-modules.md
+++ b/porting-modules.md
@@ -187,7 +187,7 @@ Define each image with a version tag to ensure that the images used are consiste
 #### Pull-through registry
 
 For most images, we use a pull-through cache in our AWS account to speed up transfers and image pulls.
-To simplify management, we keep the source image name in the containers.config files, and then prepend the pull-through registry address defined in the `params.pullthrough_registry` parameter, if it exists, using the `Utils.pullthroughContainer()` function within the module container directive.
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
👍

PR #194, opened 2025-09-30
Add batch nmve profile

### LINKED ISSUE #123: Add profiles for NVMe test batch environment
Following https://github.com/AlexsLemonade/OpenScPCA-infra/issues/113, we will need to create an alternative `batch-nvme` profile to allow launching the workflow into the NVMe-based AWS Batch environment.

jashapiro:
Stacked on #193 
closes #123 

Here I am adding a separate batch profile that uses the nvme AWS batch queue set up in https://github.com/AlexsLemonade/OpenScPCA-infra/pull/141.

I have tested it with a simulated run at https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/3cFH66yk1uVKMZ/v2/tasks, and the main takeaway is that it seems to remove all of the Docker staging issues that we had seen. A comparison with the previous simulated data run is in progress (https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/325O3uXhnQTncb/v2/tasks), but I have already seen the Docker staging issues which were plaguing us before. Some of that could be fixed by increasing IOPS with the existing EBS, but I think the NVMe instances are probably the way to go.

I will update this with links to full run results as well, which is where I might expect to see execution time differences.

I'm filing this as a draft for now, but I expect that I may want to simply replace the existing batch config with this one if it holds up for the full test. (I removed the old batch config already.) 
Let me know your thoughts on that question!


allyhawkins (commented):
This approach seems fine to me and I'm good with you removing the old batch config. I don't think there's a reason we need to keep it around.

jashapiro:
A couple of metrics from Seqera, though these are not rigorous testing:

| run type | wall time | CPU hours | estimated cost | 
| -- | -- | --:| --:|
| [sim data ebs](https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/325O3uXhnQTncb/v2/metrics) | 4h 9m | 236.4 | $15.86 |
| [sim data nvme](https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/3cFH66yk1uVKMZ/v2/metrics) | 2h 21 | 44.0 |  $4.63 |
| [full data ebs](https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/2KVGYiY6Qud5pR/v2/metrics) | 12h 58 | 517.6 | $71.31 |
| [full data nvme](https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/3zrCpHVRW28OeO/v2/metrics) | 11h 37m | 115.2 | $59.40 |

So, I think that says that the NVMe change does make for some decent savings, even if the exact data might not be perfectly reliable. For example, the CPU time measure must be off; not sure if it is maybe not multiplying by cpu count or something? But nonetheless, the NVMe drives do essentially remove the Docker staging errors that we were getting.

allyhawkins (approved):
LGTM

PR #202, opened 2026-03-18
Account for existing scimilarity results

allyhawkins:
This fixes the errors seen in https://cloud.seqera.io/orgs/CCDL/workspaces/OpenScPCA/watch/5YRvrcbETGAInN/logs. 

The most recent data release now has existing scimilarity annotations in the objects which was causing a joining error with the scimilarity results in this workflow. I decided that if they are in the object already we should use those results rather than the ones output from the `cell-type-scimilarity` module. Theoretically they should be the same, but I think it makes sense to use what's on the Portal. 

To account for that, I check if they are already present and only add them in if they aren't present. I then made sure we account for the join columns correctly, so I had to split up one of the `if` statements, but I believe this should solve our joining issue. Let me know if you disagree with using the existing annotations and would rather me remove the columns and use the results being output from the module here.

### DIFF
--- a/modules/cell-type-consensus/resources/usr/bin/assign-consensus-celltypes.R
+++ b/modules/cell-type-consensus/resources/usr/bin/assign-consensus-celltypes.R
@@ -89,6 +89,9 @@ option_list <- list(
 # Parse options
 opt <- parse_args(OptionParser(option_list = option_list))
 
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM

PR #203, opened 2026-03-24
Fix typo in ontology column name

allyhawkins:
Looks like I had a typo in #202 that meant we didn't replace the ontology column like intended.


jashapiro (approved):
:oops:

PR #204, opened 2026-04-30
remove entrypoints and use strict syntax

jashapiro:
Inspired by https://github.com/AlexsLemonade/ews-nf/issues/202, I decided to go ahead and update this workflow for strict syntax. 

The changes I made are to create a new default workflow which calls each of the workflows that were previously used as entrypoints. Now there is a `--workflow` parameter that serves the same function. The main workflow does still handle the parameter validation and intial setup before passing it to the sub-workflows, so now that code isn't duplicated at least.


jashapiro:
Apparently we need closures around publishDir statements.

allyhawkins (approved):
LGTM

PR #206, opened 2026-05-05
Prep for v0.1.8

allyhawkins:
Towards #205 

Here I'm just adding in the updated version and a changelog entry for the new release for the paper.


jashapiro (commented):
LGTM. Realized we don't have a real changelog for scpca-nf, but we should make one...

allyhawkins:
> LGTM. Realized we don't have a real changelog for scpca-nf, but we should make one...

@jashapiro I agree! But also, I think you meant to approve?

jashapiro (approved):

